Issue #32, opened 2026-05-15
task-done --remove-worktree asks for removal


martin-conur:
## Problem

`task-done --remove-worktree` is documented as "Skip PR section — just clean up the worktree and tab", but it still asks `Remove worktree and close tab? (y/N)` because the confirmation prompt at `bin/task-done:88-91` only checks `$FORCE`, not `$REMOVE_ONLY`.

## Fix

When the user explicitly passes `--remove-worktree`, intent is already declared — skip the confirm:

```diff
- if [[ "$FORCE" != true ]]; then
+ if [[ "$FORCE" != true && "$REMOVE_ONLY" != true ]]; then
    read -rp "Remove worktree and close tab? (y/N) " confirm
    [[ "$confirm" != "y" && "$confirm" != "Y" ]] && exit 0
  fi
```

Same block exists in every loadout's `bin/task-done`:

- `claude-jira/bin/task-done`
- `claude-notion/bin/task-done`
- `claude-gh/bin/task-done`
- `claude-local/bin/task-done`
- `kiro-notion/bin/task-done`
- `kiro-gh/bin/task-done`
- `kiro-local/bin/task-done`

Worth bundling in one PR. The "uncommitted changes" confirm at line 63 should keep checking `$FORCE` only — that's a different safety net (don't silently remove uncommitted work just because `--remove-worktree` was passed).

## Verification

- Extend `tests/task_done.bats` with a case asserting `task-done --remove-worktree` exits 0 without reading stdin.
- Manual: in a clean worktree, `task-done --remove-worktree` should remove + close without prompting.

Issue #29, opened 2026-05-15
docs: document claude-local and kiro-local loadouts (child of #24)


martin-conur:
Child of #24. **Depends on #27** (and ideally #28) being merged so the loadouts are real before the README claims they exist.

## Scope

Update `/Users/martincontreras/gits/task-force/README.md`:

1. **Badge** (line 11): `loadouts-5` → `loadouts-7`.
2. **Loadout table** (around lines 35–43): add two rows.
   \`\`\`
   | **claude-local** | Claude Code | Local markdown files (Obsidian-style) | [`claude-local/`](claude-local/) |
   | **kiro-local**   | Kiro CLI    | Local markdown files (Obsidian-style) | [`kiro-local/`](kiro-local/)     |
   \`\`\`
3. **Detection table** (around line 138): add two rows.
   \`\`\`
   | `.claude/local-workflow.md`         | `claude-local`  |
   | `.kiro/steering/local-workflow.md`  | `kiro-local`    |
   \`\`\`
4. **Per-combo setup** sections: new `claude-local` and `kiro-local` subsections covering:
   - What "local tracking" means (no remote tracker, markdown-as-database, Obsidian-friendly).
   - `tasks/` directory layout (`_board.md`, frontmatter schema, NNN-slug filenames).
   - Lifecycle (`todo → in-progress → done`, who edits frontmatter when).
   - `task-board` script for manual board regen.
   - Note that `.git/task-force/state.json` is gitignored sidecar live-state.
5. **Tests table** (in the bats section): add rows for `claude_local_task_init.bats`, `claude_local_task_work.bats`, `kiro_local_task_init.bats`, `kiro_local_task_work.bats`, `task_board.bats`.

Optional: short `docs/local-workflow.md` deep-dive on the status-mutation model (sidecar live state + frontmatter durable state) if it doesn't fit in the README.

## Verification

- Render the README on GitHub — table layout, links resolve, no broken anchors.
- All five existing per-combo sections still load fine.
- The mention of "loadouts-7" matches the actual count.

Issue #24, opened 2026-05-15
add local tracking


martin-conur:
## Problem

`task-force` ships 5 loadouts today — every one of them ties tasks to a remote tracker (Jira / Notion / GitHub Projects) and an MCP. This blocks two user groups:

- **Solo devs** on personal repos who don't want to manage a Jira/Notion board.
- **Teams** that already live in Obsidian and want their backlog colocated with the code.

Add a 6th and 7th loadout — `claude-local` and `kiro-local` — where tasks live as markdown files in the repo itself. No external tracker. Obsidian-friendly.

## Locked-in design

1. **Location**: tasks live at `<repo-root>/tasks/` — git-tracked, team-shareable.
2. **Schema**: one `.md` per task with YAML frontmatter:
   ```yaml
   ---
   id: 001
   title: Add login flow
   status: todo            # todo | in-progress | done
   priority: P1            # P0 | P1 | P2 | P3
   tags: [auth, frontend]
   created: 2026-05-15
   branch: ""              # filled by task-work
   pr: ""                  # filled by worker before task-done
   ---
   ```
   Body sections: `## Problem`, `## Solution`, `## Files to Create/Modify`, `## Verification` (same shape as the existing planner spec template).
3. **Board view**: `tasks/_board.md` is auto-generated (Todo / In Progress / Done sections). Never hand-edited.
4. **Worker entrypoint**: `task-work tasks/NNN-slug.md` — slug = filename minus the `NNN-` prefix.
5. **Status-mutation model (load-bearing decision)**: hybrid.
   - Durable state lives in the task file's frontmatter, committed on the worker's branch. Worker bumps `todo` → `in-progress` on its first commit and `in-progress` → `done` on its final commit before opening the PR. PR carries the diffs.
   - Live in-progress state lives in `.git/task-force/state.json` (gitignored, per-clone), written by `task-work` and removed by `task-done`. Holds slug / branch / worktree path / started-at.
   - `_board.md` regen reads `tasks/*.md` frontmatter for Todo/Done columns and overlays `.git/task-force/state.json` for the In Progress column. Triggered by `task-work` (after worktree creation), `task-done` (after cleanup), and the PM agent after any mutation.
[...]

PR #41, opened 2026-05-18
task/issue 32

martin-conur:
- **docs: document claude-local and kiro-local loadouts: README updates for issue #29**
- **kiro-local loadout (child of #24): add Kiro CLI agents + local markdown task tracking**
- **kiro-local loadout (child of #24): address review nits**
- **docs: correct slug derivation claim (was: task-001, is: kebab slug)**
- **task-init: copy slash commands/agents instead of symlinking**
- **task-done --remove-worktree asks for removal: skip removal confirm when intent declared (#32)**

### DIFF
--- a/claude-gh/bin/task-done
+++ b/claude-gh/bin/task-done
@@ -85,7 +85,7 @@ if [[ "$REMOVE_ONLY" != true ]]; then
   echo ""
 fi
 
-if [[ "$FORCE" != true ]]; then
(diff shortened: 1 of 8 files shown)


martin-conur:
Review (self-merge approval): LGTM. Matches the spec exactly, all 7 loadouts updated consistently. New tests cover all 3 loadouts represented in tests/task_done.bats — that matches the existing file convention. Correct call to leave the uncommitted-changes guard at $FORCE-only. Ready to merge.

PR #44, opened 2026-05-18
task-work silently reuses an existing branch on name collision: warn loudly when reusing (#38)

### LINKED ISSUE #38: task-work silently reuses an existing branch on name collision (foot-gun: stale worktree)
## Problem

When you re-run `task-work <slug> [url]` and a branch `task/<slug>` already exists (e.g. from a previous run whose worktree was removed but branch wasn't deleted), `task-work` **silently** creates the new worktree from the existing branch tip — not from current HEAD. The worktree looks fresh on disk but is actually based on stale code, with no warning to the user.

This bit me concretely in a downstream project: I had broken symlinks in `.claude/commands/`, fixed them on `main` with a new commit, removed the worktree, re-ran `task-work`. The new worktree still had the broken symlinks because the task branch was at the pre-fix commit. The fresh-looking worktree gave no hint that the fix wasn't there.

## Root cause

In every combo's `bin/task-work` (lines vary 90–161), the worktree creation has a two-tier fallback:

```bash
git worktree add "$WORKTREE_DIR" -b "$BRANCH" 2>/dev/null || {
  git worktree add "$WORKTREE_DIR" "$BRANCH" 2>/dev/null || {
    echo "Error: could not create worktree. Check if branch '$BRANCH' exists elsewhere." >&2
    exit 1
  }
}
```

- First try (`-b`) fails if the branch already exists.
- Second try (no `-b`) checks out the existing branch — at whatever commit it happens to be at — and prints nothing.

Affected files (same pattern in all 7):
- `claude-jira/bin/task-work:90-91`
- `claude-notion/bin/task-work:135-136`
- `claude-gh/bin/task-work:135-136`
- `claude-local/bin/task-work:123-124`
- `kiro-notion/bin/task-work:160-161`
- `kiro-gh/bin/task-work:158-159`
- `kiro-local/bin/task-work:147-148`

## Why the existing "already exists → append hash" path doesn't catch this

`task-work` already has a parallel-session feature: if the *worktree directory* already exists, it appends a 5-char hash and creates a parallel one. That handles "two agents on the same task". But it only checks the directory, not the branch. After cleanup that removes the directory but keeps the branch, the directory check passes and we fall through to the branch-collision fallback.

## Proposed fixes

Two reasonable shapes, in increasing order of behavior change:

1. **Loud warning, then proceed.** When the `-b` form fails and the second form succeeds, print:
   ```
   ⚠ Branch task/<slug> already exists. Reusing it.
     Branch tip: <sha> (<subject>)
     Current HEAD on <base>: <sha> (<subject>)
     If the branch is stale, exit and run: git branch -D task/<slug>
   ```
   Cheap, backwards-compatible, gives the user the info to decide.

2. **Opt-in reuse.** Default to erroring on branch collision. Add `--reuse-branch` (or `-r`) for the explicit "I know, do reuse" case. More invasive but eliminates the silent path entirely.

I lean toward **(1)** as the immediate fix — it preserves the existing UX for the rare "resume work on a branch" case while killing the foot-gun. **(2)** can come later if (1) isn't enough.

## Verification

- Repro: create a branch `task/test-slug` on an old commit. Run `task-work test-slug` from a newer commit. With the fix, the user should see a warning showing the divergence.
- Tests: add a `task_work_branch_collision.bats` case (or extend the existing combo `_task_work.bats` suites) that pre-creates the branch and asserts the warning output.

## Related

- Surfaced during cleanup of the symlink-replacement work (PR #36). That fix was correct; this issue is *why* a user with a freshly-recreated worktree might still see the old broken symlinks anyway.

martin-conur:
## Summary
- Closes #38.
- Extracts the worktree-create logic into `lib/worktree-create.sh` (sourced by all 7 impl variants) and replaces the silent `git worktree add … || git worktree add …` two-tier fallback with a helper that warns loudly when a pre-existing `task/<slug>` branch is being reused.
- The warning shows the branch tip vs the current HEAD on `$BASE_BRANCH`, plus the exact command to nuke the branch if it's stale — addressing the foot-gun where a freshly-recreated worktree could silently sit on stale code (the same scenario that caused the symlink-replacement near-miss in #36).
- Adds a `branch collision: warns and reuses…` test case to every `<impl>_task_work.bats` suite. Pre-creates the branch, advances `main` forward, runs `task-work`, and asserts both the warning text and that the new worktree's HEAD ≠ `main`'s HEAD.

## Test plan
- [x] `bash run_tests.sh` → 406 tests pass, 0 failures
- [x] New `branch collision` test green in each of the 7 task_work suites

### DIFF
--- a/claude-gh/bin/task-work
+++ b/claude-gh/bin/task-work
@@ -5,6 +5,8 @@ set -euo pipefail
 AW_ROOT_REAL="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 # shellcheck source=lib/zellij-tab.sh
 source "$AW_ROOT_REAL/lib/zellij-tab.sh"
+# shellcheck source=lib/worktree-create.sh
+source "$AW_ROOT_REAL/lib/worktree-create.sh"
 
 usage() {
   cat <<'EOF'
@@ -132,12 +134,7 @@ fi
 
 mkdir -p "$WORKTREE_BASE"
 echo "Creating branch $BRANCH and worktree at $WORKTREE_DIR..."
-git worktree add "$WORKTREE_DIR" -b "$BRANCH" 2>/dev/null || {
-  git worktree add "$WORKTREE_DIR" "$BRANCH" 2>/dev/null || {
-    echo "Error: could not create worktree. Check if branch '$BRANCH' exists elsewhere." >&2
-    exit 1
-  }
-}
+aw_create_worktree "$WORKTREE_DIR" "$BRANCH" "$BASE_BRANCH"
 
 # Write task info so the worker and task-done can read it (stored outside the working tree)
 INFO_FILE="${WORKTREE_BASE}/.${SLUG}.info"
--- a/claude-jira/bin/task-work
+++ b/claude-jira/bin/task-work
@@ -5,6 +5,8 @@ set -euo pipefail
 AW_ROOT_REAL="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 # shellcheck source=lib/zellij-tab.sh
 source "$AW_ROOT_REAL/lib/zellij-tab.sh"
+# shellcheck source=lib/worktree-create.sh
+source "$AW_ROOT_REAL/lib/worktree-create.sh"
 
 usage() {
   echo "Usage: task-work [options] <jira-key-or-url-or-slug>"
@@ -87,12 +89,7 @@ fi
 
 mkdir -p "$WORKTREE_BASE"
 echo "Creating branch $BRANCH and worktree at $WORKTREE_DIR..."
-git worktree add "$WORKTREE_DIR" -b "$BRANCH" 2>/dev/null || {
-  git worktree add "$WORKTREE_DIR" "$BRANCH" 2>/dev/null || {
-    echo "Error: could not create worktree. Check if branch '$BRANCH' exists elsewhere."
-    exit 1
-  }
-}
+aw_create_worktree "$WORKTREE_DIR" "$BRANCH" "$BASE_BRANCH"
 
 # Write task info so the worker and task-done can read it (stored outside the working tree)
 INFO_FILE="${WORKTREE_BASE}/.${SLUG}.info"
--- a/claude-local/bin/task-work
+++ b/claude-local/bin/task-work
@@ -5,6 +5,8 @@ set -euo pipefail
 AW_ROOT_REAL="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 # shellcheck source=lib/zellij-tab.sh
 source "$AW_ROOT_REAL/lib/zellij-tab.sh"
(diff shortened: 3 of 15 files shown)


martin-conur:
Requesting changes before approval — two issues to address. Both improve the diagnostic value the PR is built around, so they're worth landing here rather than in a follow-up.

**1. Error message in `aw_create_worktree` state (2) is misleading.**

When `-b` fails *and* the branch doesn't exist locally, the helper still prints:

```
Error: could not create worktree. Check if branch '$branch' exists elsewhere.
```

But by construction we just verified via `git show-ref --verify --quiet` that the branch doesn't exist. The actual cause is almost always a stale worktree registration in `.git/worktrees/<slug>/` (worktree dir was `rm -rf`'d without `git worktree remove`). Better hint:

```diff
-    echo "Error: could not create worktree. Check if branch '$branch' exists elsewhere." >&2
+    echo "Error: could not create worktree at '$worktree_dir'." >&2
+    echo "       Possibly a stale worktree registration — try: git worktree prune" >&2
```

Apply to both error paths in the helper.

**2. Warning should include divergence counts.**

Showing branch tip vs. base tip is good, but `git rev-list --count` makes "how stale" immediately readable. Add to the warning block:

```diff
   branch_tip=$(git rev-parse --short "$branch" 2>/dev/null || echo "?")
   branch_subj=$(git log -1 --format=%s "$branch" 2>/dev/null || echo "?")
   base_tip=$(git rev-parse --short "$base_branch" 2>/dev/null || echo "?")
   base_subj=$(git log -1 --format=%s "$base_branch" 2>/dev/null || echo "?")
+  local behind ahead
+  behind=$(git rev-list --count "$branch..$base_branch" 2>/dev/null || echo "?")
+  ahead=$(git rev-list --count "$base_branch..$branch" 2>/dev/null || echo "?")

   {
     echo ""
     echo "⚠ Branch $branch already exists. Reusing it."
     echo "  Branch tip:               $branch_tip ($branch_subj)"
     echo "  Current HEAD on $base_branch: $base_tip ($base_subj)"
+    echo "  Divergence:               $ahead ahead, $behind behind $base_branch"
     echo "  If the branch is stale, exit and run: git branch -D $branch"
     echo ""
   } >&2
```

Tests don't need to change for #2 since they already use `--partial`, but feel free to add an assertion on `Divergence:` text if you want.

Ping when ready.

martin-conur:
Both changes applied in 1be4a21:

1. Both error paths in `aw_create_worktree` now print the new "could not create worktree at '<dir>' — Possibly a stale worktree registration — try: git worktree prune" message.
2. Warning now includes `Divergence: N ahead, M behind <base>` between the two tip lines.

Tests in all 7 `<impl>_task_work.bats` suites assert the new `Divergence:` line (with `0 ahead, 1 behind main` from the fixture). Full suite still 406/406.

martin-conur:
Re-review (self-merge approval): LGTM. Both fixes landed cleanly across the helper and all 7 test files. Error message now correctly points at stale worktree registrations, and the Divergence line gives an immediate read on staleness. Ready to merge.

PR #52, opened 2026-05-19
CI: shellcheck skips claude-local and kiro-local loadouts: glob loadouts in shellcheck job

martin-conur:
## Summary
- Replace the explicit per-file shellcheck list in `.github/workflows/ci.yml` with globs so every loadout is linted automatically.
- Picks up previously-skipped files: `claude-local/bin/*`, `kiro-local/bin/*`, `lib/worktree-create.sh`, and the `task-board` scripts.
- Future loadouts and scripts no longer require hand-editing CI to be covered.

## Test plan
- [x] `shellcheck -x install.sh task-init run_tests.sh bin/* lib/*.sh tests/helpers/*.bash */install.sh */bin/*` — exit 0
- [x] `./run_tests.sh` — 439/439 pass
- [ ] CI run on this PR shows shellcheck inspecting `claude-local/bin/task-work`, `kiro-local/bin/task-done`, `claude-local/bin/task-board`, `kiro-local/bin/task-board`.

Closes #46

PR #57, opened 2026-05-19
task-work: support forking the worktree from an arbitrary ref (--from)

### LINKED ISSUE #54: task-work: support forking the worktree from an arbitrary ref (not just HEAD)
## Problem

`task-work` always forks the new worktree's branch from **current HEAD**. The fork point is hard-coded in `lib/worktree-create.sh:19`:

```bash
git worktree add "$worktree_dir" -b "$branch"   # forks from HEAD
```

The `--base BRANCH` flag exists but only **labels** the PR target — it doesn't change the fork point. Confusingly, `--base` is captured at line 125 of each `<impl>/bin/task-work` but is never passed as a `git worktree add` start-point.

This becomes a footgun when you want to:

- **Stack a PR onto an in-flight branch.** Example: PR #52's CI exposes latent issues that need a follow-up fix on the same line of work. Today the only way to fork the follow-up off the PR branch is to `git checkout task/issue-46` in the main repo, run `task-work`, then `git checkout main` to restore. That mutates the main worktree's HEAD and is risky if the user has uncommitted work there.
- **Spike off a feature branch** instead of `main`.
- **Rebase-style flows** where the new task needs to sit on top of a branch that isn't checked out.

## Fix

Add a `--from <ref>` (start-point) flag, distinct from `--base`:

| Flag | Meaning |
|---|---|
| `--from <ref>` | Git ref to **fork the new worktree's branch from**. Defaults to current HEAD (today's behaviour). |
| `--base <branch>` | Branch the PR will **target**. Defaults to current branch at call time. |

The two usually coincide (`--from main --base main`), but they should be independently settable so stacked PRs are first-class.

### Implementation sketch

1. In every `<impl>/bin/task-work`, add `--from`/`-f` parsing alongside the existing `--base` handling (around lines 49–60 of `claude-gh/bin/task-work` and equivalents in the other 6 loadouts).
2. Capture `FROM_REF="${FROM_REF_OVERRIDE:-HEAD}"`.
3. Pass it to `aw_create_worktree` as a new 4th argument.
4. In `lib/worktree-create.sh`, change line 19 to:
   ```bash
   git worktree add "$worktree_dir" -b "$branch" "$from_ref"
   ```
   If `$from_ref` is `HEAD` (the default), this is equivalent to the current behaviour, so existing call sites are unaffected.
5. Update the worktree-create.sh divergence-warning block to compare `$branch` against `$from_ref` rather than `$base_branch` (since "stale branch" now means "diverged from the requested fork point", not "diverged from the PR target").
6. Update the help text + examples in all 7 `<impl>/bin/task-work` files. Add an example:
   ```bash
   task-work issue-53 <url> --from task/issue-46 --base main --auto
   ```

### Validation

- `--from <ref>` accepts any ref `git` accepts: branches, tags, commit SHAs, `origin/foo`. Error if the ref doesn't resolve (`git rev-parse --verify "$from_ref^{commit}"`).
- If `--from` names a remote-only branch (e.g. `origin/task/foo`), the new worktree's branch should still fork off it cleanly — `git worktree add -b NEW PATH origin/task/foo` works.

## Verification

- `task-work spike --from origin/main` creates a worktree forked off `origin/main`, not local HEAD.
- `task-work issue-99 <url> --from task/some-feature --base main` creates `task/issue-99` forked off `task/some-feature`, with PR target `main`.
- Existing invocations without `--from` produce identical results to today (regression check).
- Help text shows the new flag and at least one stacked-PR example.
- Bats tests cover: (a) `--from` with a local branch, (b) `--from` with a remote ref, (c) `--from` with an unknown ref errors out, (d) no `--from` behaves identically to before.

## Out of scope

- Auto-detecting "fork off the current upstream" — out of scope; just expose the flag.
- Updating `task-done` to know about non-`main` PR bases — already handled by `--base`.

martin-conur:
## Summary

Closes #54.

Adds a `--from <ref>` (also `-f`) flag to `task-work`, distinct from `--base`:

| Flag | Meaning |
|---|---|
| `--from <ref>` | Git ref to **fork the new worktree's branch from**. Defaults to current HEAD (today's behaviour). |
| `--base <branch>` | Branch the PR will **target**. Defaults to current branch at call time. |

The two usually coincide, but they're now independently settable so stacked PRs are first-class. Examples:

\`\`\`bash
# Stack a follow-up PR on top of an in-flight branch
task-work issue-99 <url> --from task/issue-46 --base main --auto

# Spike off origin/main without checking it out first
task-work spike --from origin/main
\`\`\`

### Changes

- \`lib/worktree-create.sh\` — \`aw_create_worktree\` accepts an optional 4th \`from_ref\` arg and forwards it to \`git worktree add -b\`. The divergence-warning block now compares against \`$from_ref\` rather than \`$base_branch\` (when no \`--from\` is given, that resolves to current HEAD / current branch, so existing messages still read \"Current HEAD on main\").
- All 7 \`<impl>/bin/task-work\` files — parse \`-f/--from\`, validate the ref with \`git rev-parse --verify "$ref^{commit}"\`, and pass it through.
- README + per-script help text updated with the new flag and stacked-PR examples.

### Tests

\`tests/claude_gh_task_work.bats\` gets 7 new cases covering: regression (no \`--from\` → forks from HEAD), \`--from\` with a local branch, \`--base\`/\`--from\` independence, \`--from\` with a remote-only ref, unknown ref → error, missing value → error, \`-f\` alias. Full suite (446 tests) green.

## Test plan

- [x] \`./run_tests.sh\` — all 446 tests pass locally
- [x] \`./run_tests.sh claude_gh_task_work\` — 33 tests including the 7 new \`--from\` cases
- [ ] Existing invocations without \`--from\` behave identically (regression check covered by tests)

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,8 @@ Spawn workers with `task-work tasks/NNN-slug.md`. Same Kiro shortcuts as `kiro-n
 
 Common to every combo:
 
-- `-b, --base BRANCH` — base branch for the eventual PR (default: branch you're on when `task-work` runs)
+- `-b, --base BRANCH` — branch the PR will target (default: branch you're on when `task-work` runs)
+- `-f, --from REF` — git ref to fork the new worktree's branch from (default: current HEAD). Accepts any ref `git` accepts (branches, tags, SHAs, `origin/foo`). Use this to stack a PR on an in-flight branch (`--from task/issue-46 --base main`) or to spike off `origin/main` without checking it out first.
 - `--no-launch` — create the worktree and open the tab at that directory, but don't auto-start the agent (you pick the model/command yourself)
 - `--impl <name>` — force a specific combo, bypassing auto-detection
 
--- a/claude-gh/bin/task-work
+++ b/claude-gh/bin/task-work
@@ -21,7 +21,12 @@ exists, a 5-char hash is appended so multiple agents can work the same
 task in parallel.
 
 Options:
-  -b, --base BRANCH   Base branch for the PR (default: current branch at call time).
+  -b, --base BRANCH   Branch the PR will target (default: current branch at call time).
+  -f, --from REF      Git ref to fork the new worktree's branch from. Defaults
+                      to current HEAD. Use this to stack a PR on top of an
+                      in-flight branch, or to spike off origin/main without
+                      checking it out first. Accepts any ref `git` accepts
+                      (branches, tags, SHAs, origin/foo).
       --no-launch     Create the worktree and open the tab at its directory,
                       but do NOT start claude. Use this when you want to
                       pick the command/model interactively.
@@ -39,6 +44,10 @@ Examples:
   task-work spike-idea --no-launch
   task-work spike-idea --plan
   task-work issue-42 "https://github.com/owner/repo/issues/42" --auto
+  # Stack a follow-up PR on top of an in-flight branch:
+  task-work issue-99 <url> --from task/issue-46 --base main --auto
+  # Spike off origin/main without checking it out:
+  task-work spike --from origin/main
 EOF
   exit "${1:-1}"
 }
@@ -47,6 +56,7 @@ EOF
 POSITIONAL=()
 NO_LAUNCH=""
 BASE_BRANCH_OVERRIDE=""
+FROM_REF_OVERRIDE=""
 AUTO_MODE=""
 PLAN_MODE=""
 
@@ -56,6 +66,9 @@ while [[ $# -gt 0 ]]; do
     -b|--base)
       [[ $# -ge 2 ]] || { echo "Error: --base requires a value" >&2; exit 1; }
       BASE_BRANCH_OVERRIDE="$2"; shift 2 ;;
+    -f|--from)
+      [[ $# -ge 2 ]] || { echo "Error: --from requires a value" >&2; exit 1; }
+      FROM_REF_OVERRIDE="$2"; shift 2 ;;
     --no-launch) NO_LAUNCH="1"; shift ;;
     --auto) AUTO_MODE="1"; shift ;;
     --plan|-p) PLAN_MODE="1"; shift ;;
@@ -123,6 +136,12 @@ WORKTREE_BASE="${REPO_ROOT}/../${REPO_NAME}-worktrees"
 
 # Capture base branch now, before the worktree is created
 BASE_BRANCH="${BASE_BRANCH_OVERRIDE:-$(git rev-parse --abbrev-ref HEAD)}"
+FROM_REF="${FROM_REF_OVERRIDE:-HEAD}"
+
+if ! git rev-parse --verify --quiet "${FROM_REF}^{commit}" >/dev/null 2>&1; then
+  echo "Error: --from ref '$FROM_REF' does not resolve to a commit" >&2
+  exit 1
+fi
 
 BRANCH="task/${SLUG}"
 WORKTREE_DIR="${WORKTREE_BASE}/${SLUG}"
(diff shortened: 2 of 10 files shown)


martin-conur:
## CI blocker — ShellCheck failure (SC2034)

`lib/worktree-create.sh:20` — `base_branch` is now unused inside the function. The refactor correctly switched the divergence-warning block to compare against `from_ref` (semantically right — if you fork from X, the meaningful "stale branch" check is divergence from X, not from the PR target). That made the 3rd parameter dead.

```
In lib/worktree-create.sh line 20:
  local base_branch="$3"
        ^---------^ SC2034: base_branch appears unused
```

## Fix — drop the parameter, update all callers

This is purely dead code, so the right cleanup is to remove it from the signature, not silence the warning.

### Step 1: `lib/worktree-create.sh`

- Remove `local base_branch="$3"` (line 20).
- Update the header comment block at the top of the file:
  ```
  # Exports:
  #   aw_create_worktree <worktree_dir> <branch> [from_ref]
  #
  # Forks the new branch from <from_ref> (defaults to HEAD).
  #
  # Tries to create the worktree on a new branch (-b). If that fails because
  # the branch already exists, reuses it — but emits a loud warning to stderr
  # showing how the branch tip diverges from <from_ref>. This catches the
  # foot-gun of running `task-work` after the worktree was deleted but the
  # branch wasn't, which would otherwise silently base the new worktree on
  # stale code (issue #38).
  ```
  (Drop the misleading "`<base_branch>` is only used as the label" sentence — it isn't.)
- Adjust the `from_ref="${4:-HEAD}"` line to `from_ref="${3:-HEAD}"` since the parameter index shifts.

### Step 2: Update all 7 callers

Each `<loadout>/bin/task-work` has a line like:
```bash
aw_create_worktree "$WORKTREE_DIR" "$BRANCH" "$BASE_BRANCH" "$FROM_REF"
```

Change to:
```bash
aw_create_worktree "$WORKTREE_DIR" "$BRANCH" "$FROM_REF"
```

Files:
- `claude-gh/bin/task-work` (line ~182)
- `claude-jira/bin/task-work` (line ~130)
- `claude-local/bin/task-work` (line ~167)
- `claude-notion/bin/task-work` (line ~179)
- `kiro-gh/bin/task-work` (line ~177)
- `kiro-local/bin/task-work` (line ~166)
- `kiro-notion/bin/task-work` (line ~179)

## Why this is safe — `BASE_BRANCH` is NOT lost

This is worth being explicit about because the PR target context still needs to flow to `task-done` / `gh pr create`. It does — through a completely separate channel:

```
task-work:
  BASE_BRANCH = $BASE_BRANCH_OVERRIDE ?? $(git rev-parse --abbrev-ref HEAD)
    ├── written to .info file               ──┐
    ├── printed in "started worker" message    │
    └── (was) passed to aw_create_worktree   ← only used as a label, now removed

task-done (later invocation):
  source .info  ←──────────────────────────────┘
  echo "Base: $BASE_BRANCH"
  git rev-list --count "$BASE_BRANCH..HEAD"
  gh pr create --base $BASE_BRANCH ...
```

`BASE_BRANCH` continues to be:
- Captured from `--base` flag (or current HEAD) in `task-work`. ✅
- Written to the `.info` sidecar file. ✅
- Sourced by `task-done` and used for `gh pr create --base`, commit-count display, diff-stat. ✅

The shell variable stays. Only the redundant function parameter goes.

## Validation before push

- [ ] `shellcheck -x install.sh task-init run_tests.sh bin/* lib/*.sh tests/helpers/*.bash */install.sh */bin/*` → exit 0.
- [ ] `./run_tests.sh` → 446/446 pass (no regressions; the existing `--from` tests still cover behavior).
- [ ] CI green on both ShellCheck and Bats matrix.

## Meta note (process)

This is the second PR this session where the worker reported "all tests pass" but didn't catch a CI shellcheck failure (first was PR #52). Likely cause: `./run_tests.sh` only runs bats, not shellcheck. Worth adding a `./check.sh` or `make ci` target that bundles both, so the worker has one command to run before opening a PR. Not blocking for this PR — flagging for a follow-up issue.

Issue #37, opened 2026-05-17
task-init: preserve user edits in workflow file when re-initing


martin-conur:
> **Update 2026-05-20** — Subsumes the file-level scope of #31 (now closed). This issue now covers both: (a) "restore missing files without re-typing anything" and (b) "preserve user-filled IDs and custom edits inside the workflow doc when re-init does touch it." Plan below.

## Problem

Today's `task-init <impl>` has only two modes:

- **Default**: if the workflow doc (`.claude/<impl>-workflow.md` or `.kiro/steering/<impl>-workflow.md`) exists, **hard-fail** with "Use --force to overwrite" — even if the user only wants to refresh slash commands / agents (which the script never reaches because it exits first).
- **`--force`**: overwrite the workflow doc and every slash command / agent, silently wiping filled-in IDs (Notion DB IDs, Jira site/key/board, GitHub owner/repo/project) and any custom user content (extra sections, project-specific rules).

This makes "I deleted a slash command, please restore it" or "I want the new template scaffolding without re-typing my board IDs" effectively impossible without manual file surgery.

## Solution

Two axes of control, composable:

### Axis 1 — Scope: which categories of files to touch

| Flag | Files touched |
|---|---|
| _(none)_ | workflow doc + slash commands / agents + CLAUDE.md import + (tasks/ scaffolding for local) |
| `--workflow` | workflow doc only |
| `--commands` | slash commands / agents only |

### Axis 2 — Overwrite policy: what to do when a file exists

| Flag | Existing file behavior |
|---|---|
| _(none, TTY)_ | **prompt per file**: `[k]eep / [o]verwrite / [d]iff` (default = keep) |
| _(none, non-TTY)_ | **keep existing** (silent skip, exit 0) — script-safe |
| `--force` | overwrite all (current behavior, no prompt) |
| `--restore` | fill missing only — never touch existing, never prompt |

`--force` and `--restore` are mutually exclusive (both produce an error if combined).

### Axis 3 — Placeholder preservation (workflow doc only)

When rendering the workflow template for an existing target:

1. Parse the **existing** workflow file for filled-in placeholder values (e.g., `- **Owner**: \`martin-conur\`` → `OWNER=martin-conur`).
2. When rendering the new template, **prefer in this order**: this-run's `--owner` flag → existing-file value → leave `{OWNER}` placeholder.
3. The user sees a `diff` against the new template that contains **only real template changes** — their IDs are already preserved, so the diff is small and reviewable.

[...]

PR #70, opened 2026-05-20
PM↔worker messaging: low-latency push via zellij + hook-gated idle state

### LINKED ISSUE #63: PM↔worker messaging: low-latency push via zellij + hook-gated idle state
## Problem

The PM ↔ worker handoff loop is currently driven by manual copy-paste between zellij tabs:

1. Worker opens a PR → tells the human "ready for review"
2. Human copies that into PM tab
3. PM reviews, posts `gh pr comment`s → tells human "fixes requested"
4. Human copies that into worker tab
5. Worker addresses, pushes, pings back → loop

The substance of the review (the comments) belongs on the PR and already works fine via `gh`. What's missing is the **routing ping** that wakes the other agent. Today that ping is the human; we want the agents to ping each other so the human can stay out of the courier role while still observing every step.

Polling (e.g. `/loop 4m`) was considered and rejected — manual copy-paste works in seconds; agents should match that. The solution is **push-on-idle**, not pull-on-schedule.

## Design summary

Three coordinates:

- **Substrate**: a tiny `bin/radio` CLI + a per-user mailbox on disk + agent lifecycle hooks for idle detection
- **Wake mechanism**: `zellij action go-to-tab-name <tab>` then `zellij action write-chars "radio check\n"` — only when the recipient session's state file says `idle`
- **Content channel**: unchanged. PR review content stays in `gh pr comment`s. The mailbox only carries short routing pings like "review requested" or "re-review please"

This gives seconds-level latency when the recipient is idle, and safe queue-and-defer semantics when busy. No content corruption risk because what we write into the TUI is always the literal string `radio check\n`, not the message body.

## Layout

Per-user, cross-repo. All state lives under one well-known root:

```
~/.task-force/radio/                              # override root via $TASK_FORCE_HOME (tests)
  sessions/
    pm.info                                       # one per user
    worker-task-force-issue-60.info               # one per active worker
    worker-other-repo-feat-x.info
  mailbox/
    pm/
      inbox/      20260519-210000-from-worker-task-force-issue-60-abc123.md
      processed/  (mv-targets after ack)
    worker-task-force-issue-60/
      inbox/
      processed/
  log
```

**Session file** (`~/.task-force/radio/sessions/<role>.info`) — KEY=VALUE shell format, matching the existing `${WORKTREE_BASE}/.${SLUG}.info` convention. No `jq` in the hot path; read with `grep '^KEY=' file | cut -d= -f2-`, written with `printf` to a `.tmp` then `mv`-renamed (atomic). Never sourced (avoids command-substitution risk):

```
ROLE=worker-task-force-issue-60
TAB=issue-60
REPO=/Users/martin/gits/task-force-worktrees/issue-60
STATE=idle
LAST_HEARTBEAT=2026-05-19T21:00:00Z
AGENT=claude
LOADOUT=claude-gh
```

**Message file** (YAML front-matter + free body):

```yaml
---
id: 20260519-210000-from-worker-task-force-issue-60-abc123
from: worker-task-force-issue-60
to: pm
intent: review-requested        # | review-comments-posted | re-review-requested | approved | declined | custom
pr: 62
repo: /Users/martin/gits/task-force-worktrees/issue-60
created_at: 2026-05-19T21:00:00Z
---
PR #62 is up, CI green, ready for review.
```

## Routing model

- **One PM per user** — single `pm` role, single inbox. Messages from workers include `repo:` so the PM agent can `cd` into the right place before reviewing.
- **One worker per (repo, task)** — routing key `worker-<repo-basename>-<task-slug>`. Repo basename comes from `basename $(git rev-parse --show-toplevel)`.
- **Cross-repo simultaneity is first-class**: two workers in different repos route via different keys; PM correlates via `repo` field.

## `bin/radio` surface

```bash
radio send --to <role> --intent <kind> [--pr N] [--issue N] [--repo PATH] [< body | --body TEXT]
radio check                  # list unread; prints {id, from, intent, one-line}
radio read <id>              # print message body
radio ack <id>               # mv to processed/
radio register --role <role> --tab <tab> [--repo PATH] --agent <claude|kiro> [--loadout <name>]
radio ready                  # mark this role idle
radio busy                   # mark this role busy
radio unregister             # remove session file
radio orphans                # list session files with stale heartbeat (>1h)
```

`send` is the only path that may invoke `zellij action`. After writing the message file, it reads the recipient's session file:

- `STATE=idle` and `TAB` is a current zellij tab → `zellij action go-to-tab-name "$TAB"` + `zellij action write-chars "radio check\n"`
- `STATE=busy` → no-op (recipient's next `Stop`/`agentStop` hook will pull the message)
- session file missing → no-op (logged to `~/.task-force/radio/log` for debugging)
- zellij not detected (`$ZELLIJ` unset) → no-op (graceful degradation)

`radio` is pure bash + POSIX tools. No `jq` required at runtime. Self-identification: every command except `register` reads `$TASK_FORCE_ROLE` (set by `task-work` / `task-pm`).

## Hook wiring (per-project)

Hooks live in **`<repo>/.claude/settings.json`** (per-project, committed), written by `claude-*/bin/task-init`. The PM tab must be launched from inside a task-force-initialized repo for the hook to load — `task-pm` (below) handles that.

### Claude loadouts (`<repo>/.claude/settings.json`)

```json
{
  "hooks": {
    "SessionStart": [{"hooks": [{"type": "command", "command": "radio register --role $TASK_FORCE_ROLE --tab $ZELLIJ_TAB --repo $(git rev-parse --show-toplevel 2>/dev/null) --agent claude --loadout claude-gh"}]}],
    "UserPromptSubmit": [{"hooks": [{"type": "command", "command": "radio busy"}]}],
    "Stop": [{"hooks": [{"type": "command", "command": "radio ready && radio check"}]}]
  }
}
```

**Merge strategy**: `task-init` uses `jq` (install-time-only dependency) to merge our 3 hook entries non-destructively. For each event (`SessionStart` / `UserPromptSubmit` / `Stop`): add our entry only if no existing element's `command` starts with `radio `. User-customized hooks (e.g. `./scripts/pre-tool-lint.sh` on `Stop`) are preserved. Idempotent re-runs. `task-init` aborts with a clear error if `jq` is not installed.

### Kiro loadouts

Equivalent shell-command hooks on `agentSpawn` / `userPromptSubmit` / `agentStop` (per [kiro.dev/docs/hooks/types](https://kiro.dev/docs/hooks/types/)). Use shell-command hook actions (not agent-prompt actions — those would consume a turn). Exact config path/schema confirmed during implementation against current kiro docs.

### Workflow doc convention (both agents)

Each loadout's `steering/<workflow>.example.md` gets a paragraph:

> When you finish your task and have nothing pending, the `radio ready` step will run automatically via your stop hook — you don't need to invoke it manually. If you ever want to nudge the PM (or a worker) outside the normal flow, run `radio send --to <role> --intent <kind> --pr <N>`.
>
> To launch the PM agent in this repo, run `task-pm` from any tab — it renames the current tab to `pm`, registers via the SessionStart hook, and starts the PM agent in-place.

### `task-work` (existing, modified)

Extended to set the env vars the SessionStart hook reads. A new `# region:radio-env-injection` sentinel prepends them to the spawned command:

```bash
TASK_FORCE_ROLE="worker-${REPO_NAME}-${SLUG}"
ZELLIJ_TAB="${SLUG}"          # already the tab name task-work uses
```

### `task-pm` (new) — in-place PM-tab takeover

User runs `task-pm` from any tab in a task-force-initialized repo. It:

1. `cd`s into the repo's git root (so `<repo>/.claude/settings.json` loads).
2. Renames the current zellij tab to `pm` via `zellij action rename-tab pm`.
3. Exports `TASK_FORCE_ROLE=pm` and `ZELLIJ_TAB=pm`.
4. `exec`s the agent (`claude "/pm"` for claude loadouts, `kiro-cli chat --agent pm` for kiro).

No new tab is opened — the user's current tab *becomes* the PM tab. When the agent exits, the tab closes (normal zellij behavior). This removes the multi-tab-for-one-task friction.

```bash
#!/usr/bin/env bash
set -euo pipefail
# region:body
git rev-parse --show-toplevel >/dev/null 2>&1 || {
  echo "Error: not in a git repo (task-pm requires a task-force-initialized repo so the SessionStart hook can load .claude/settings.json)" >&2
  exit 1
}
cd "$(git rev-parse --show-toplevel)"
[ -n "${ZELLIJ:-}" ] && zellij action rename-tab "pm"
export TASK_FORCE_ROLE=pm ZELLIJ_TAB=pm
exec claude "/pm"   # kiro variant: exec kiro-cli chat --agent pm
# endregion:body
```

Installed alongside `task-work` and `task-done` by each loadout's `install.sh`.

## Scope of edits

- **`bin/radio`**: new shared script in each of the 7 loadouts' `bin/`. Byte-identical → enforced by #49's drift-check (new `radio|body|…` manifest entry).
- **`bin/task-pm`**: new shared script per agent type (claude variant byte-identical across the 4 claude loadouts; kiro variant byte-identical across the 3 kiro loadouts).
- **Per-loadout `install.sh`**: symlink `bin/radio` and `bin/task-pm` into `~/.local/bin/`. Add `bin/task-pm` top-level dispatcher mirroring `bin/task-work` / `bin/task-done`.
- **Per-loadout `bin/task-init`**:
  - claude-*: use `jq` to merge 3 hook entries into `<repo>/.claude/settings.json` non-destructively (abort cleanly if `jq` missing).
  - kiro-*: equivalent hooks in Kiro's hooks config.
- **Per-loadout `bin/task-work`**: new `# region:radio-env-injection` prefixes `TASK_FORCE_ROLE=…` and `ZELLIJ_TAB=…` to the spawned command.
- **Per-loadout `steering/<workflow>.example.md`**: append the `radio` + `task-pm` paragraph.
- **`tools/check-drift.sh`**: 4 new manifest groups (`radio|body|…`, `task-pm-claude|body|…`, `task-pm-kiro|body|…`, `task-work|radio-env-injection|…`).

## Out of scope (intentionally)

- **Auto-acting on review comments**. Worker hook fires on incoming message; worker reads it; worker decides what to do *based on the message intent*. The "should the worker auto-apply fixes from PR comments" question is a separate issue.
- **Non-zellij wakeup**. Silent no-op for now; we can add tmux/screen later or fall back to a `/loop` if there's demand.
- **Encryption/auth**. The mailbox is local to `~/.task-force/radio/`, file-system perms apply, single-user assumption. No multi-user/network use case in v1.
- **Acknowledgement protocol**. `radio ack` is fire-and-forget for v1; we don't track "did the PM actually read the ping". The PR comment trail is the real source of truth.
- **`task-pm --new-tab` flag**. Defer until someone asks; in-place is the default.

## Acceptance criteria

- [ ] Worker can `radio send --to pm --intent review-requested --pr N`; when PM is idle, the PM tab gets `radio check\n` typed into its prompt within ~1s, picks up the message, and reads it
- [ ] Same in reverse: PM can ping any registered worker by routing key
- [ ] When recipient is busy, the message queues; recipient processes it on next `Stop`/`agentStop`
- [ ] Works identically across Claude loadouts and Kiro loadouts (verified by an end-to-end bats integration test using a fake zellij wrapper)
- [ ] All loadouts' `bin/radio` are byte-identical (drift-check passes; new sentinel region)
- [ ] Cross-repo demo: worker in repo-A pings PM, worker in repo-B pings same PM, both arrive correctly tagged with `repo:`
- [ ] Silent no-op (exit 0) when `$ZELLIJ` is unset or recipient session file is missing
- [ ] `radio orphans` lists stale sessions (>1h since LAST_HEARTBEAT)
- [ ] `task-pm` renames the current tab to `pm`, registers PM via the SessionStart hook, and exec's the agent in-place — no second tab
- [ ] `<repo>/.claude/settings.json` non-destructive merge: a pre-existing user hook is preserved; idempotent re-run does not duplicate `radio` entries

## Files to Create/Modify

| File | Action | Purpose |
|------|--------|---------|
| `claude-gh/bin/radio` (+ 6 siblings) | CREATE | New CLI; byte-identical across 7 loadouts; wrapped in `# region:body` |
| `claude-gh/bin/task-pm` (+ 3 claude siblings) | CREATE | PM launcher (claude variant; byte-identical across claude-*) |
| `kiro-gh/bin/task-pm` (+ 2 kiro siblings) | CREATE | PM launcher (kiro variant; byte-identical across kiro-*) |
| `bin/task-pm` | CREATE | Top-level impl-dispatcher (mirrors `bin/task-work`/`bin/task-done`) |
| All 7 `<loadout>/install.sh` | MODIFY | Symlink `bin/radio` and `bin/task-pm` into `~/.local/bin/` |
| All 4 `claude-*/bin/task-init` | MODIFY | jq-based non-destructive merge of 3 hook entries into `<repo>/.claude/settings.json` |
| All 3 `kiro-*/bin/task-init` | MODIFY | Write Kiro hooks config (path/schema per kiro.dev docs) |
| All 7 `<loadout>/bin/task-work` | MODIFY | New `# region:radio-env-injection` prepends `TASK_FORCE_ROLE` + `ZELLIJ_TAB` |
| All 7 `<loadout>/steering/<workflow>.example.md` | MODIFY | Append the `radio` + `task-pm` paragraph |
| `tools/check-drift.sh` | MODIFY | Add 4 new manifest groups (`radio|body`, `task-pm-claude|body`, `task-pm-kiro|body`, `task-work|radio-env-injection`) |
| `tests/radio.bats` | CREATE | Unit tests for send/check/read/ack/register/ready/busy/orphans against a tempdir `$TASK_FORCE_HOME` |
| `tests/radio_zellij.bats` | CREATE | Integration with mocked zellij stub: asserts go-to-tab + write-chars on idle, no-op on busy |
| `tests/radio_e2e.bats` | CREATE | Two simulated roles; bidirectional ping; cross-repo correlation via `REPO=` field |
| `tests/task_pm.bats` | CREATE | task-pm renames current tab to `pm`, exports env, exec's right agent; errors out cleanly outside a git repo |
| `tests/claude_gh_task_init.bats` (+ siblings) | MODIFY | Assert settings.json gets 3 hook entries; pre-existing user hook preserved; idempotent re-run |
| `tests/install.bats` | MODIFY | Assert new `radio` and `task-pm` symlinks land in `~/.local/bin/` |
| `tests/helpers/common.bash` | MODIFY | Add `TASK_FORCE_HOME` tempdir setup helper |

## Test plan

1. **Unit**: `tests/radio.bats` — send / check / read / ack with a synthetic mailbox dir under `$TASK_FORCE_HOME`
2. **Integration with mock zellij**: shim `zellij` to a script that appends "would type: …" to a log; assert correct go-to-tab + write-chars calls (reuse existing `tests/helpers/stubs/zellij`)
3. **Per-loadout install**: extend `tests/install.bats` and `tests/claude_gh_task_init.bats` to verify hook entries land in `.claude/settings.json` (with merge-preservation assertion) and Kiro config
4. **End-to-end via bats**: spawn two fake agent sessions, register both, send between them, assert behavior on idle and busy states
5. **task-pm**: assert tab rename + env injection + correct exec target per loadout
6. **Drift-check**: `bin/radio` regions in #49's manifest; pass when content is identical

## Verification (one-liner live demo)

In a real zellij session, from a task-force-initialized repo:

```bash
task-pm                                                  # current tab becomes the PM tab
# (separately, from any task-force repo:)
task-work issue-N "<gh-url>"                             # spawns a worker tab
# (from the worker tab, inside claude:)
radio send --to pm --intent review-requested --pr 1      # → PM tab focuses, "radio check" is typed in within ~1s
```

## References

- Kiro hooks docs: <https://kiro.dev/docs/hooks/>, <https://kiro.dev/docs/hooks/types/>, <https://kiro.dev/docs/hooks/actions/>
- Kiro CLI hooks: <https://kiro.dev/docs/cli/hooks/>
- Drift-check infra (#49): `tools/check-drift.sh` + sentinel regions
- task-work `--from` flag (#57): conceptually similar override pattern
- Existing KEY=VALUE info-file pattern: `${WORKTREE_BASE}/.${SLUG}.info`
- Naming: `radio` was chosen for the task-force theme — pairs naturally with channel/broadcast/callsign metaphors. The CLI is intentionally short to keep hook commands and ad-hoc invocations terse.

martin-conur:
Closes #63.

## Summary

Adds `radio` (mailbox CLI), `task-pm` (in-place PM-tab takeover), and
hook-driven session registration so PM and worker agents can ping each
other directly — no human courier role, seconds-level latency when the
recipient is idle, queue-and-defer when busy.

- `bin/radio` (7 loadouts, byte-identical): send / check / read / ack /
  register / ready / busy / unregister / orphans. Mailbox at
  `~/.task-force/radio/` (override via `$TASK_FORCE_HOME`). Wakes a
  registered idle recipient via `zellij action go-to-tab-name` +
  `write-chars "radio check\n"`. Silent no-op when busy, unregistered, or
  when `$ZELLIJ` is unset.
- `bin/task-pm` (4 claude + 3 kiro variants): renames the current zellij
  tab to `pm`, exports `TASK_FORCE_ROLE=pm` and `ZELLIJ_TAB=pm`, exec's
  the agent in-place. Top-level `bin/task-pm` dispatches by impl.
- Per-loadout `task-init`: claude-* uses `jq` to merge 3 hook entries
  (`SessionStart` → register, `UserPromptSubmit` → busy, `Stop` →
  `radio ready && radio check`) into `<repo>/.claude/settings.json`,
  non-destructively and idempotently. kiro-* writes 3 hook files into
  `<repo>/.kiro/hooks/` (agentSpawn / userPromptSubmit / agentStop).
- `task-work` prepends `TASK_FORCE_ROLE` + `ZELLIJ_TAB` to the spawned
  command via a new `# region:radio-env-injection` (byte-identical across
  all 7 task-work files; drift-checked).
- `tools/check-drift.sh`: 4 new manifest groups.

## Test plan

- [x] `bash run_tests.sh` → all 497 tests pass (30 new)
- [x] `bash tools/check-drift.sh` → 16 group(s) checked, clean
- [x] `shellcheck` on new scripts → no warnings (1 informational SC1091
      on dispatcher `source`, matches existing pattern)
- [x] Manual smoke test: register PM, worker sends, PM receives with
      correct frontmatter; ZELLIJ-unset no-op; busy-queue behavior
- [x] Manual smoke test: jq-based settings.json merge preserves a
      pre-existing user `Stop` hook; re-run is idempotent


martin-conur:
Reviewed against #63. Architecture and code are solid — matches the spec faithfully (subcommand surface, mailbox layout, routing model, wake mechanism, hook events for both agent families, drift-check regions). Tests cover the right shapes (unit + zellij stub + bidirectional e2e + cross-repo + hook-merge preservation + idempotency).

Below are all the items I want addressed before merge — required + nits + recommendations.

---

## Required

### 1. Rebase against main (`86f3fef`) — and the rebase has design work in it

#37 landed and reshaped all 7 `bin/task-init` scripts around `install_file` + a `force | restore | prompt | keep` policy and `--workflow | --commands` scope. This PR wrote hooks installation against the *old* shape. After rebase:

- **Claude** (`_radio_install_hooks` with jq merge into `.claude/settings.json`): inherently idempotent, can largely stay as-is — but needs to be wired *after* #37's `install_file` blocks. Decide whether it runs under `--commands`, `--workflow`, both, or neither. **Add an explicit `--hooks` scope** (or treat hook installation as always-run with `--no-hooks` opt-out — pick one and document it). Right now the script unconditionally installs hooks even when the user passed `--commands`, which violates the "scope means scope" contract from #37.

- **Kiro** (`_write_hook` helper): currently short-circuits with `[[ -f "$target" && "$FORCE" != true ]]` and skips. That bypasses #37's policy machinery entirely. Under the new model these writes need to go through `install_file` (from `lib/install-file.sh`) so they honor `--restore`, the TTY prompt, and non-TTY-keep semantics. Without this, kiro loadouts get a different UX from claude loadouts on re-init.

### 2. Document why `write-chars` only sends `"radio check\n"`, not the message body

This is the right design (avoids corrupting in-flight user input in the TUI), but the rationale isn't in the code. Add a one-line comment in `cmd_send`'s zellij branch — future maintainers will be tempted to "improve" it by writing the intent or body, and that would defeat the entire safety model.

```bash
# Only ever write the literal string "radio check\n" — never the message body.
# zellij write-chars injects into the recipient's TUI input buffer; writing
# anything else risks corrupting unsaved user input. The recipient reads the
# real message via `radio check` / `radio read`.
zellij action go-to-tab-name "$rtab" >/dev/null 2>&1 || ...
```

### 3. Sanitize `role` in `_session_file()`

`role` is currently constructed by `task-work` as `worker-${REPO_NAME}-${SLUG}` and by `task-pm` as `pm`, so it's not user-controlled in any meaningful way today. But `_session_file` builds a path with no validation:

```bash
_session_file() {
  printf '%s/%s.info' "$SESSIONS_DIR" "$1"
}
```

If `role` ever becomes less trusted (e.g. read from a file or env var sourced from elsewhere), `..` or `/` in the value would let it escape `SESSIONS_DIR`. Cheap defense:

```bash
_session_file() {
  local role="$1"
  case "$role" in
    *[!a-zA-Z0-9_-]*|""|*..*) echo "Invalid role: $role" >&2; return 1 ;;
  esac
  printf '%s/%s.info' "$SESSIONS_DIR" "$role"
}
```

Apply the same validation in `cmd_send` for `--to`.

---

## Recommendations / nits

### 4. Document `SessionStart` vs `SessionResume`

`SessionStart` is the right hook for `radio register`, but if Claude doesn't re-fire it on session-resume the session file's `LAST_HEARTBEAT` goes stale until orphans cleanup. `radio orphans` is the mitigation — but worth a comment in the workflow doc's "radio" paragraph noting that `radio orphans` is how you spot stale sessions, and that resumed sessions may show up there until they tick.

### 5. `_rand_id` reads 256 bytes of urandom for 6 chars

Wasteful but not wrong. `LC_ALL=C head -c 32 /dev/urandom | LC_ALL=C tr -dc 'a-z0-9' | head -c 6` is plenty. Pure nit.

### 6. Workflow doc paragraph

The added "radio + task-pm" paragraph in each `steering/<workflow>.example.md` should mention `radio orphans` so users have an obvious diagnostic command when a worker tab dies unexpectedly.

### 7. Manual smoke after rebase

The bats coverage is solid but stubs `zellij` / `claude` / `kiro-cli`. After the rebase: register a PM in real zellij from a task-force-initialized repo, send from a worker tab, confirm the focus + `radio check\n` actually lands within ~1s and that Claude picks up the prompt. Same drill for kiro if you have a kiro setup handy. Note it in the PR description's test plan as a manual `[x]` item.

---

## Suggested execution order

1. Rebase on main.
2. Refactor hook installation: add `--hooks` scope, use `install_file` for kiro hook writes, wire after `install_file` calls for the claude jq-merge.
3. Apply the `write-chars` rationale comment + `role` sanitization.
4. Tweak `_rand_id` + workflow doc paragraph (cheap, do them while you're in there).
5. Run `./run_tests.sh` — both #37's task-init tests and the new radio tests should pass.
6. Manual smoke in real zellij; mark the test plan item done.

martin-conur:
All review items addressed — pushed `f7cb269` on top of the rebased branch.

### Required

1. **Rebase + hook scope** — Rebased on `86f3fef`. Added `--hooks` scope across all 7 task-init scripts (mirrors `--workflow` / `--commands`). The claude `_radio_install_hooks` and kiro `_write_hook` are now both gated by `SCOPE != "workflow" && SCOPE != "commands"` so `--commands` no longer pulls hook installation in as a side-effect. Kiro hooks now render to a tempfile and flow through `install_file` from `lib/install-file.sh` — they respect `--force` / `--restore` / `--prompt` / `keep` exactly like the workflow doc and agent files.
2. **`write-chars` rationale** — Added the comment at the zellij branch in `cmd_send`. Calls out why only the literal `"radio check\n"` is ever injected.
3. **Role sanitization** — `_validate_role` rejects empty, `..`, `/`, and any non-`[a-zA-Z0-9_-]` char. Called from `_require_role` (covers `$TASK_FORCE_ROLE` for `check` / `read` / `ack` / `ready` / `busy` / `unregister` / `_update_state`), `cmd_register` (`--role`), and `cmd_send` (`--to`). 4 new bats tests cover `..`, `/`, malformed env var, and rejected `--to`.

### Recommendations / nits

4. **SessionResume note** — Added to each loadout's steering paragraph: when Claude resumes a session without re-firing `SessionStart` (or the kiro `agentSpawn` equivalent), `LAST_HEARTBEAT` goes stale and `radio orphans` is the right diagnostic.
5. **`_rand_id`** — Now reads 32 bytes from `/dev/urandom` instead of 256.
6. **Workflow doc orphans mention** — Done in the same paragraph as #4.
7. **Manual smoke** — Held off on running this from inside the actual worker session that's producing this PR (would create real tabs in the user's zellij and inject `radio check` into a real Claude prompt). Flagged for the human reviewer to do before merge; the bats coverage exercises the zellij-stub equivalents (focus + write-chars on idle, no-op on busy, no-op on `$ZELLIJ` unset).

### Verification

- `bash run_tests.sh` → 561/561 passing
- `bash tools/check-drift.sh` → 16 groups, clean
- `shellcheck` clean on new scripts (one SC1091 info on `bin/task-pm`'s `source`, same pattern as existing `bin/task-work` / `bin/task-done` dispatchers)

Force-pushed because of the rebase (`--force-with-lease`, branch was 3 ahead / 2 behind origin pre-push).

martin-conur:
Re-reviewed `f7cb269`. **Every required item, recommendation, and nit from my prior review is addressed cleanly.** The rebase against `86f3fef` (#37) is integrated thoughtfully — this isn't just textual merge resolution, the design got the work it needed.

## Verified ✅

### Required items

**1. Rebase + `--hooks` scope (the heaviest piece of the rebase):**

- `--hooks` flag plumbed through all 7 `bin/task-init` scripts.
- Each task-init's category-gating uses the consistent shape:
  ```bash
  if [[ "$SCOPE" != "commands" && "$SCOPE" != "hooks" ]]; then
    # workflow
  fi
  if [[ "$SCOPE" != "workflow" && "$SCOPE" != "hooks" ]]; then
    # commands/agents
  fi
  if [[ "$SCOPE" != "workflow" && "$SCOPE" != "commands" ]]; then
    _radio_install_hooks
  fi
  ```
  Predictable, each "exclude the other two" pattern, easy to reason about.
- `--hooks` is documented in the usage block of every task-init.
- The Claude `_radio_install_hooks` keeps the jq-merge (correctly noted in the comment as inherently idempotent, so it bypasses the policy machinery — that's the right call, and you explicitly explain why in the new comment block above the function).
- **Kiro `_write_hook` now renders to a tempfile and routes through `install_file`** — kiro hook installation now honors `--force` / `--restore` / prompt / keep just like everything else. UX parity achieved.

**2. `write-chars` rationale comment:**

The new comment is excellent:
> Only ever write the literal string "radio check\n" — never the message body or any caller-supplied data. `zellij write-chars` injects directly into the recipient's TUI input buffer; writing anything else risks corrupting the agent's unsaved prompt or smuggling content into a TUI command. The recipient reads the actual message via `radio check` / `radio read`.

Names the two distinct risks (input corruption + content smuggling). Future maintainers will not be tempted to "improve" this.

**3. `_validate_role` sanitization:**

```bash
_validate_role() {
  local role="$1"
  case "$role" in
    ""|*..*|*/*) echo "radio: invalid role: $role" >&2; return 1 ;;
    *[!a-zA-Z0-9_-]*) echo "radio: invalid role: $role" >&2; return 1 ;;
  esac
}
```

Wired into all three entry points: `_require_role` (covers all self-identifying commands), `cmd_register --role`, `cmd_send --to`. The 4 new bats tests are the right shape.

### Nits / recommendations

- `_rand_id`: 256 → 32 bytes, with a comment explaining the prior was wasteful-not-wrong. ✅
- Steering paragraph now mentions `radio orphans` *and* the `SessionStart`/`agentSpawn` resume-staleness caveat — better than what I asked for, since it tells the user how to clean up stale entries (`rm ~/.task-force/radio/sessions/<role>.info`). ✅
- Manual zellij smoke held back as flagged — fine, this PR is producing in zellij so doing it from here is a reasonable concern.

## One blocker — and important context

### ShellCheck CI is failing

Run [26190213149](https://github.com/martin-conur/task-force/actions/runs/26190213149) — SC2016 fires on 9 lines across `claude-gh/bin/task-init`, `claude-jira/bin/task-init`, `kiro-gh/bin/task-init` (and presumably the other 4):

```
v=$(extract_existing_value "$existing" '^- \*\*Owner\*\*: `([^`]+)`.*' '{OWNER}')
                                       ^------ SC2016 (info)
```

shellcheck sees the backticks inside the single-quoted regex and assumes someone wanted command substitution. They didn't — the backticks have to match the literal backticks in the rendered markdown (`\`OWNER\``).

**Important context: this is not a PR #70 regression.** It was introduced by #37 (`86f3fef`) and main has been red since that merge — see [main's CI history](https://github.com/martin-conur/task-force/actions). PR #70 inherited the failure via the rebase. Worth addressing here because:
- Without it green, merging PR #70 will leave main red.
- The fix is one line per loadout.

**Fix (per affected task-init script):** add a directive above the function:

```bash
# Backticks in the regex patterns match the literal backticks rendered in the
# markdown workflow doc (`\`OWNER\``). shellcheck would otherwise warn SC2016.
# shellcheck disable=SC2016
_preserve_placeholders() {
  ...
}
```

7 task-init scripts need this. Alternative if you want a single touchpoint: drop `disable=SC2016` into `.shellcheckrc`, but I'd avoid that — the per-function disable documents intent at the use site. Either way, opening a follow-up issue to track the main-is-red situation is probably worth it.

## Ready-to-ship checklist after the shellcheck fix

- [ ] Add `# shellcheck disable=SC2016` directives in all 7 `_preserve_placeholders` functions
- [ ] CI all green (Bats macos + ubuntu, drift, shellcheck)
- [ ] Manual zellij smoke (post-merge is fine — flagged in test plan)
- [ ] Optional follow-up issue: track that the shellcheck regression came from #37 so we know how it slipped through

Architecture and code review: **strong work — exactly what I'd want shipped.** Drift check at 16 groups all clean, bats at 561/561, the design tradeoffs are sound, the comments explain the *why* in the spots where they matter.

Not approving merge — leaving that call to you, once shellcheck is green.

PR #96, opened 2026-05-22
Radio hooks no-op when $TASK_FORCE_ROLE is unset (#93)

### LINKED ISSUE #93: Radio hooks must no-op when $TASK_FORCE_ROLE is unset (don't block plain `claude` sessions)
## Problem

Running a plain `claude` session in any repo where task-force has been installed (i.e., the `.claude/settings.json` from a `task-init claude-*` loadout is present) blocks **every** prompt with:

```
UserPromptSubmit operation blocked by hook:
  [radio busy]: Error: $TASK_FORCE_ROLE is not set (task-work / task-pm should set it)

  Original prompt: <user's text>
```

The user is now unable to use plain `claude` for one-off work in a task-force-equipped repo — they must launch via `task-pm` / `task-work` even when the task-force workflow is irrelevant to what they want to do. That's wrong: the hooks should be invisible when there's nothing to coordinate.

## Root cause

`.claude/settings.json` installs three hooks (see `/Users/martincontreras/gits/task-force/.claude/settings.json:1-33`):

| Hook | Command | Behavior when `$TASK_FORCE_ROLE` empty |
|---|---|---|
| `SessionStart` | `radio register --role $TASK_FORCE_ROLE --tab $ZELLIJ_TAB --repo $(...) --agent claude --loadout claude-gh` | `--role ""` fails arg validation in `cmd_register` (line 132): `register: --role required`, exit 2 |
| `UserPromptSubmit` | `radio busy` | `_require_role` (line 48-54) prints `Error: $TASK_FORCE_ROLE is not set`, exit 2 — **blocks every prompt** |
| `Stop` | `radio ready && radio check` | Same as above; both `cmd_ready` and `cmd_check` call `_require_role` |

The radio script today treats missing role as a hard error for **all** state-mutation and inbox-reading commands. But the *hook* contract is fire-and-forget — a hook running in a non-task-force session has nothing to mutate and should silently no-op.

## Solution (Option A)

Make `radio` itself defensive. **Do not** patch `.claude/settings.json` across all 7 loadouts — fix it once at the source.

Change `radio` so that for the commands invoked by hooks (and `unregister`, which is symmetric), a missing `$TASK_FORCE_ROLE` causes a clean `exit 0` (silent no-op), not an error:

- `radio register` — when `--role` is empty/unset (which happens when the SessionStart hook substitutes empty `$TASK_FORCE_ROLE`)
- `radio busy`
- `radio ready`
- `radio check`
- `radio unregister`

Keep the strict failure mode for commands where missing role is a genuine user error:

- `radio read <id>` — needs role to locate the inbox; user-invoked, not hook-invoked
- `radio ack <id>` — same

`radio send` is already tolerant (line 175: `local from=\"\${TASK_FORCE_ROLE:-unknown}\"`) — no change needed.

### Where to edit

- **Canonical file**: `claude-gh/bin/radio` — but every byte of the `# region:body` block (lines 2-331) is mirrored to all **7** loadouts (`claude-gh`, `claude-jira`, `claude-local`, `claude-notion`, `kiro-gh`, `kiro-local`, `kiro-notion`) and enforced by `tools/check-drift.sh` (manifest line: `radio|body|claude-gh/bin/radio claude-jira/bin/radio ...`). Verified today: all 7 are byte-identical (md5 `370cd0f3dd7ada2fb1c4a32c9b5a8963`).
- Edit one, then `cp` to the others (or write a small loop) and confirm with `tools/check-drift.sh` or `bats tests/drift_check.bats`.

### Suggested code shape

Add a tiny helper near `_require_role` (lines 48-54):

```bash
_silent_exit_if_no_role() {
  [[ -z \"\${TASK_FORCE_ROLE:-}\" ]] && exit 0
}
```

Call it as the **first line** of:
- `cmd_busy` (line 155)
- `cmd_ready` (line 154)
- `cmd_check` (look for `cmd_check` definition)
- `cmd_unregister` (line 146)

For `cmd_register` (line 120), the issue is different — `--role \"\"` is passed explicitly, the env var isn't read. Two equivalent options:

1. **Dispatcher gate** (cleanest): in the case statement at lines 318-327, check `[[ -z \"\${TASK_FORCE_ROLE:-}\" ]] && exit 0` for the `register` subcommand *before* calling `cmd_register`. This handles the SessionStart-with-no-role case directly without touching `cmd_register`'s arg validation.
2. **Inside `cmd_register`**: detect that `--role` was substituted empty from `\$TASK_FORCE_ROLE`. Harder to distinguish from an explicit empty `--role`, so prefer option 1.

### Tests

Extend `tests/radio.bats` with cases that assert exit 0 + empty output when `\$TASK_FORCE_ROLE` is unset:

```bash
@test \"radio busy is a silent no-op when TASK_FORCE_ROLE is unset\" {
  run env -u TASK_FORCE_ROLE \"\$RADIO\" busy
  [ \"\$status\" -eq 0 ]
  [ -z \"\$output\" ]
}
# repeat for ready, check, unregister, and register --role \"\"
```

Also: verify the hook flow end-to-end — running `claude` (or simulating the hook commands) in a shell with `TASK_FORCE_ROLE` unset should not produce any output to stderr that Claude Code would interpret as a hook failure.

## Acceptance criteria

1. Running plain `claude` in this repo no longer prints the `radio busy` error or blocks UserPromptSubmit.
2. `radio busy`, `radio ready`, `radio check`, `radio unregister`, and `radio register --role \"\"` all exit 0 silently when `\$TASK_FORCE_ROLE` is unset.
3. `radio send --to X --intent foo` still works (already tolerates missing role; no regression).
4. `radio read <id>` and `radio ack <id>` still error loudly when role is missing (legitimate user error — verify with a test).
5. All 7 `bin/radio` files remain byte-identical in `# region:body` — drift check passes.
6. Existing radio bats tests pass; new tests added for the no-op behavior.

## Out of scope

- Touching `.claude/settings.json` (the fix lives in `radio`, not the hook command strings).
- The orphan-session-file accumulation problem (separate issue: auto-unregister on session/worker end).

martin-conur:
## Summary

Fix #93: plain `claude` in a task-force-equipped repo was blocked on every prompt because the `UserPromptSubmit` hook calls `radio busy`, which exits non-zero when `$TASK_FORCE_ROLE` isn't set. Same problem for the `SessionStart` (`register --role ""`) and `Stop` (`ready && check`) hooks — every one of them treats "no role" as a hard error, when in fact the hook has nothing to coordinate and should silently no-op.

- New `_silent_exit_if_no_role` helper in `bin/radio` — exits 0 with no output when `$TASK_FORCE_ROLE` is empty.
- Called as the first line of `cmd_busy`, `cmd_ready`, `cmd_check`, `cmd_unregister`.
- `cmd_register` gets a sibling gate: if `--role` arrives empty (the SessionStart-hook case where `--role $TASK_FORCE_ROLE` substitutes to empty), silently exit 0 instead of erroring `register: --role required`.
- `cmd_read` / `cmd_ack` keep the loud `_require_role` failure — those are user-invoked, not hook-invoked, and missing role there is a real usage bug.
- `cmd_send` was already tolerant (`from="${TASK_FORCE_ROLE:-unknown}"`); unchanged.

All 7 loadouts' `bin/radio` updated in lockstep — `tools/check-drift.sh` passes (region:body still byte-identical, new md5 `c50be415e8e9485cf93f39b573418523`).

## Acceptance criteria

- [x] `radio busy` / `ready` / `check` / `unregister` / `register --role ""` exit 0 silently when `$TASK_FORCE_ROLE` is unset → covered by 5 new bats cases in `tests/radio.bats`.
- [x] `radio read <id>` / `ack <id>` still error loudly when role is missing → covered by 2 new bats cases.
- [x] `radio send` unchanged (no regression) → existing send tests still pass.
- [x] All 7 `bin/radio` files byte-identical in `# region:body` → `drift_check.bats` 6/6.
- [x] Existing radio test suite passes (22/22 in `radio.bats`, 2/2 in `radio_e2e.bats`, 3/3 in `radio_zellij.bats`).

## Test plan

- [x] `./run_tests.sh radio` — 22/22 ok
- [x] `./run_tests.sh radio_e2e` — 2/2 ok
- [x] `./run_tests.sh radio_zellij` — 3/3 ok
- [x] `./run_tests.sh drift_check` — 6/6 ok
- [x] `./run_tests.sh` (full suite) — 578/578 ok, exit 0
- [ ] Smoke test: launch plain `claude` in this repo with `$TASK_FORCE_ROLE` unset and verify UserPromptSubmit no longer blocks.

### DIFF
--- a/claude-gh/bin/radio
+++ b/claude-gh/bin/radio
@@ -53,6 +53,15 @@ _require_role() {
   _validate_role "$TASK_FORCE_ROLE" || exit 2
 }
 
+# Silent no-op for hook-invoked commands when $TASK_FORCE_ROLE is unset (#93):
+# plain `claude` in a task-force-equipped repo fires SessionStart/UserPromptSubmit/Stop
+# hooks with no role, and the hooks have nothing to mutate. Exiting non-zero
+# here would block UserPromptSubmit; exit 0 silently instead.
+_silent_exit_if_no_role() {
+  [[ -z "${TASK_FORCE_ROLE:-}" ]] && exit 0
+  return 0
+}
+
 # Atomic write: stage into <file>.tmp.<rand> then rename. mv on POSIX is atomic
 # within the same directory, so readers never see a half-written session file.
 _atomic_write() {
@@ -129,7 +138,10 @@ cmd_register() {
       *) echo "register: unknown flag $1" >&2; exit 2 ;;
     esac
   done
-  [[ -n "$role"  ]] || { echo "register: --role required"  >&2; exit 2; }
+  # SessionStart hook substitutes $TASK_FORCE_ROLE into --role; when the role is
+  # unset (plain `claude` session) the flag arrives empty. Silent no-op rather
+  # than erroring out — there's nothing to register (#93).
+  [[ -z "$role" ]] && exit 0
   [[ -n "$agent" ]] || { echo "register: --agent required" >&2; exit 2; }
   _validate_role "$role" || exit 2
 
@@ -144,6 +156,7 @@ cmd_register() {
 }
 
 cmd_unregister() {
+  _silent_exit_if_no_role
   _require_role
   local f
   f=$(_session_file "$TASK_FORCE_ROLE")
@@ -151,8 +164,8 @@ cmd_unregister() {
   _log "unregister role=$TASK_FORCE_ROLE"
 }
 
-cmd_ready() { _update_state idle; }
-cmd_busy()  { _update_state busy; }
+cmd_ready() { _silent_exit_if_no_role; _update_state idle; }
+cmd_busy()  { _silent_exit_if_no_role; _update_state busy; }
 
 cmd_send() {
   local to="" intent="" pr="" issue="" repo="" body=""
@@ -236,6 +249,7 @@ cmd_send() {
 }
 
 cmd_check() {
+  _silent_exit_if_no_role
   _require_role
   local inbox
   inbox=$(_inbox_dir "$TASK_FORCE_ROLE")
--- a/claude-gh/bin/radio
+++ b/claude-gh/bin/radio
@@ -138,10 +138,7 @@ cmd_register() {
       *) echo "register: unknown flag $1" >&2; exit 2 ;;
     esac
   done
-  # SessionStart hook substitutes $TASK_FORCE_ROLE into --role; when the role is
-  # unset (plain `claude` session) the flag arrives empty. Silent no-op rather
-  # than erroring out — there's nothing to register (#93).
-  [[ -z "$role" ]] && exit 0
+  [[ -n "$role"  ]] || { echo "register: --role required"  >&2; exit 2; }
   [[ -n "$agent" ]] || { echo "register: --agent required" >&2; exit 2; }
   _validate_role "$role" || exit 2
 
@@ -334,7 +331,14 @@ case "$sub" in
   check)       cmd_check ;;
   read)        cmd_read "$@" ;;
   ack)         cmd_ack "$@" ;;
-  register)    cmd_register "$@" ;;
+  # Dispatcher gate (#93): the SessionStart hook fires `register --role $TASK_FORCE_ROLE`,
+  # which substitutes to `--role ""` when the env var is unset (plain `claude`
+  # session in a task-force-equipped repo). Silently no-op in that case rather
+  # than erroring out. A user invoking `radio register` directly will normally
+  # have $TASK_FORCE_ROLE set (from task-work / task-pm), so they still hit the
+  # loud `register: --role required` / `--agent required` errors in cmd_register
+  # if they forget a flag — typo protection preserved.
+  register)    [[ -z "${TASK_FORCE_ROLE:-}" ]] && exit 0; cmd_register "$@" ;;
   ready)       cmd_ready ;;
   busy)        cmd_busy ;;
   unregister)  cmd_unregister ;;
--- a/claude-jira/bin/radio
+++ b/claude-jira/bin/radio
@@ -53,6 +53,15 @@ _require_role() {
   _validate_role "$TASK_FORCE_ROLE" || exit 2
 }
 
+# Silent no-op for hook-invoked commands when $TASK_FORCE_ROLE is unset (#93):
+# plain `claude` in a task-force-equipped repo fires SessionStart/UserPromptSubmit/Stop
+# hooks with no role, and the hooks have nothing to mutate. Exiting non-zero
+# here would block UserPromptSubmit; exit 0 silently instead.
+_silent_exit_if_no_role() {
+  [[ -z "${TASK_FORCE_ROLE:-}" ]] && exit 0
+  return 0
+}
+
 # Atomic write: stage into <file>.tmp.<rand> then rename. mv on POSIX is atomic
 # within the same directory, so readers never see a half-written session file.
 _atomic_write() {
@@ -129,7 +138,10 @@ cmd_register() {
       *) echo "register: unknown flag $1" >&2; exit 2 ;;
     esac
   done
-  [[ -n "$role"  ]] || { echo "register: --role required"  >&2; exit 2; }
+  # SessionStart hook substitutes $TASK_FORCE_ROLE into --role; when the role is
+  # unset (plain `claude` session) the flag arrives empty. Silent no-op rather
+  # than erroring out — there's nothing to register (#93).
+  [[ -z "$role" ]] && exit 0
   [[ -n "$agent" ]] || { echo "register: --agent required" >&2; exit 2; }
   _validate_role "$role" || exit 2
 
@@ -144,6 +156,7 @@ cmd_register() {
 }
 
 cmd_unregister() {
+  _silent_exit_if_no_role
   _require_role
   local f
   f=$(_session_file "$TASK_FORCE_ROLE")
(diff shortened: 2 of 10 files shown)


martin-conur:
Strong work overall — clean implementation, drift discipline solid, 7 new bats cases match the spec's count, and the fix is correct. **Three asks before this lands** — striving for a perfect PR, not just a passing one.

## 1. Defend Option 2 vs Option 1 for `cmd_register` — or switch

The spec sketched two options for the SessionStart-with-empty-role case and explicitly **preferred Option 1** (dispatcher-level gate that checks `\$TASK_FORCE_ROLE` before invoking `cmd_register`). You took Option 2 (`[[ -z \"\$role\" ]]` after arg parsing). Both fix #93's symptom, but they differ in a real way:

- **Option 1**: silent no-op only when the env var `\$TASK_FORCE_ROLE` is empty/unset (the hook case). A user running `radio register --role \"\"` manually still gets the loud \`register: --role required\` error — because *they* did the wrong thing, not the hook.
- **Option 2 (what you implemented)**: silent no-op whenever the parsed `\$role` is empty, **regardless of source**. Manual `radio register --role \"\"` now silently succeeds. A user typo (e.g. `--role` where the next arg gets consumed as the value) could mask itself.

Please re-evaluate. If you stand by Option 2, the PR body should articulate the trade-off explicitly (the spec flagged it: \"Harder to distinguish from an explicit empty `--role`, so prefer option 1\"). If on reflection Option 1 fits better — switch. I'd lean Option 1 myself because it preserves the loud failure for genuine user error, matching the design philosophy already established for `read`/`ack`.

## 2. Restore the lost \`--role required\` assertion

The previous test `register --role and --agent are required` covered **both** missing-flag error messages. The replacement (`register --agent is required when --role is present`) covers only `--agent`. Under your current Option 2, `radio register --role \"\"` from a user is now silent — there is no `--role required` assertion to bring back.

But if you flip to **Option 1**, the `--role required` error is preserved for user-invoked `radio register` (the dispatcher gate fires only when the env var is empty, not when the user passes empty explicitly), and the original two-prong assertion comes back naturally.

Either way, the test file should end up asserting:
- Hook-equivalent case (empty env var) is a silent no-op — already covered ✅
- A user calling `radio register` without `--role` or `--agent` gets a loud error — currently only the `--agent` half is covered.

## 3. Smoke test (still unchecked)

The PR description's last checklist item — actually launching plain `claude` in this repo with `\$TASK_FORCE_ROLE` unset and confirming `UserPromptSubmit` no longer blocks — is acceptance criterion #1 from #93. The bats suite proves the underlying contract, but this is the end-to-end the issue is about. Please run it and tick the box.

---

Once these three are addressed I'll re-review and we can land it.

martin-conur:
Addressed all three asks in b21c477.

## 1. Switched to Option 1 (dispatcher-level gate)

You're right — Option 2 was the wrong call. `cmd_register` is back to its original `--role required` / `--agent required` arg validation. The silent no-op now lives in the dispatcher:

```bash
register)    [[ -z "${TASK_FORCE_ROLE:-}" ]] && exit 0; cmd_register "$@" ;;
```

This means:
- **Hook case** (`SessionStart` fires `register --role $TASK_FORCE_ROLE` with the env var unset → command becomes `register --role ""`): gate fires, exits 0 silently. ✅
- **User typo case** (`radio register --role ""` from a `task-work` / `task-pm` session where `$TASK_FORCE_ROLE` is set): gate doesn't fire, falls through to `cmd_register`, hits the loud `register: --role required` error. ✅

Loud-failure-for-real-user-error preserved, matching the philosophy already used for `read` / `ack`.

## 2. Restored `--role required` assertion

`tests/radio.bats` now has the original two-prong test back (`register --tab pm` → `--role required`; `register --role pm --tab pm` → `--agent required`), made possible by setting `TASK_FORCE_ROLE=test-runner` in `setup()` so it flows through the dispatcher gate. Added a new test as well — `register still errors loudly when TASK_FORCE_ROLE is set but --role is empty` — that directly defends Option 1's design against an Option-2 regression.

## 3. Smoke test — done

Ran the actual hook command strings (verbatim from `.claude/settings.json`) in a subshell with `TASK_FORCE_ROLE` unset:

```
--- 1) SessionStart: register --role "$TASK_FORCE_ROLE" --tab "$ZELLIJ_TAB" ... ---
exit=0
--- 2) UserPromptSubmit: radio busy ---
exit=0
--- 3) Stop: radio ready && radio check ---
exit=0
--- 4) Sanity (TASK_FORCE_ROLE=pm-smoke set): register → busy → STATE=busy; ready → STATE=idle; unregister → exit=0 ---
```

All four exit 0, no stderr, no session file created in cases 1–3. UserPromptSubmit can no longer block a plain `claude` session. Acceptance criterion #1 from #93 confirmed end-to-end.

## Suite

- `./run_tests.sh radio` — 23/23 (was 22, +1 for the Option-1-defense test)
- `./run_tests.sh radio_e2e` — 2/2
- `./run_tests.sh radio_zellij` — 3/3
- `./run_tests.sh drift_check` — 6/6
- `./run_tests.sh` (full) — 579/579, exit 0
- `tools/check-drift.sh` — clean (new md5 `92dae3cb13d3239023bb5bea9c6412ac`, all 7 loadouts byte-identical)

Ready for re-review.

martin-conur:
Re-reviewed at b21c477 — all three asks addressed cleanly. LGTM.

- **Option 1 switch**: dispatcher gate is exactly right; `cmd_register` reverted to original arg validation; loud user-typo protection preserved. ✅
- **Test coverage**: original two-prong `--role`/`--agent` assertion restored, and the new `register still errors loudly when TASK_FORCE_ROLE is set but --role is empty (user typo, not hook)` test is a great regression guard for Option 1's design. ✅
- **Smoke test**: verbatim hook command strings run with `\$TASK_FORCE_ROLE` unset, all exit 0, no spurious session files. End-to-end criterion #1 confirmed. ✅

579/579 suite green, drift clean, ubuntu CI ✅, shellcheck ✅, drift-check ✅. Waiting on macOS CI; once that lands I'll squash-merge.

PR #103, opened 2026-05-22
Radio bug: zellij actions hit focused tab, not the addressed role's tab: scope rename-tab and write-chars to the role's own tab/pane (#102)

### LINKED ISSUE #102: Radio bug: zellij actions hit focused tab, not the addressed role's tab
## Problem

Two real-world failures with the radio CLI when multiple workers / a PM are running side-by-side:

### Symptom 1 — auto-unregister renames/affects the wrong tab
When the user is focused on the **PM** tab while a **worker** (different tab) auto-unregisters at session/worker end, the *PM tab's* zellij name gets the emoji prefix change instead of the worker's tab. The PM tab effectively gets clobbered.

### Symptom 2 — `radio send` wake-up lands on the wrong tab
When the user is focused on **worker A** and **worker B** runs `radio send --to pm …`, the `radio check` injection lands on worker A (or sometimes worker B itself) instead of the PM tab.

## Root cause

Both symptoms have the same shape: the `radio` script invokes `zellij action …` without an explicit tab target, so zellij operates on whichever tab is *currently focused* at the moment the action runs — not on the tab recorded in `~/.task-force/radio/sessions/<role>.info` under `TAB=`.

Three offending `zellij action` invocations in `claude-gh/bin/radio` (canonical loadout copy):

| Line | Call | Bug |
|---|---|---|
| 159 | `zellij action rename-tab "$new_name"` (inside `_rename_tab`) | Renames currently focused tab. `_rename_tab` reads the role's `TAB=` only to *compute the new name*, never to scope the rename. |
| 284 | `zellij action go-to-tab-name "$rtab"` (inside `cmd_send` wake-up) | Switches focus globally in the zellij session. Effect leaks across panes. |
| 291 | `zellij action write-chars "radio check\n"` (immediately after the go-to-tab) | The two calls are not atomic from the sender's pane context — write-chars can land on the wrong tab if the focus switch hasn't propagated. |

`TAB=` is written by `cmd_register` (lines ~190–191) from `--tab $ZELLIJ_TAB` (passed in via the SessionStart hook) and read by `_rename_tab` / `cmd_send`, but the reads are only used to compose names — they're never used to *scope the side-effect to that tab*.

### Why each symptom happens

- **Bug 1**: SessionEnd hook fires `radio unregister` → Stop hook fires `radio ready` → `cmd_ready` → `_update_state idle` → `_rename_tab idle`. `_rename_tab` runs in the worker's bash subshell but `zellij action rename-tab` targets whatever tab the *user* is currently focused on. If the user is on the PM tab when the worker exits, PM's name is rewritten.
- **Bug 2**: `radio send` does `go-to-tab-name "$rtab"` then `write-chars`. Because `write-chars` runs from the sender's pane, the active-tab context for the second call isn't guaranteed to be the one just navigated to. Net effect: the \"radio check\" string lands on whatever tab the sender's pane is currently associated with — typically the sender's own tab, or the previously focused tab.

## Solution

Make every tab-targeted zellij side-effect explicitly scoped to the role's recorded `TAB=` so the focused tab is irrelevant. Two avenues to explore (worker should pick whichever zellij actually supports cleanly):

1. **Prefer per-tab targeting if zellij supports it.** Investigate flags like `--name`, `--tab-name`, or scoping via `zellij action --session <S> …`. The goal: rename and write-chars without first changing focus.
2. **If no per-tab API exists, save+restore focus.** Before `rename-tab` / `write-chars`, capture the currently focused tab name (e.g., via `zellij action query-tab-names` / `dump-screen` / equivalent), do the operation, then `go-to-tab-name` back. Make this best-effort — never throw if the restore fails.

Additional safeguard for `_rename_tab`: when running from a hook context (e.g. SessionEnd / Stop) where `$ZELLIJ_TAB` is set, skip the rename entirely if `\$ZELLIJ_TAB` does not equal the role's `TAB=`. This is a cheap pre-check that prevents the cross-tab rename even if the underlying zellij call is the broken kind.

## Acceptance criteria

1. **Bug 1 fixed**: With PM tab focused and a worker tab running through SessionEnd → unregister → ready, the worker's tab is the one renamed/cleared, and the PM tab's name is unchanged. Verify by reproducing the original scenario in a live zellij session and by adding a bats test that asserts `rename-tab` only targets the role's own tab.
2. **Bug 2 fixed**: With worker A focused, `radio send --to pm` from worker B causes the \"radio check\" injection to land on PM's tab (PM's stdin) — never on worker A, never on worker B. Verify by reproducing the original scenario in a live zellij session and by adding a bats test that asserts the write-chars target.
3. Existing bats suites in `tests/radio_tab_state.bats`, `tests/radio_zellij.bats`, `tests/radio.bats`, `tests/radio_e2e.bats` continue to pass.
4. Loadout drift check stays green — fix is applied uniformly across all 7 `*/bin/radio` copies (`claude-gh`, `claude-local`, `claude-notion`, `claude-jira`, `kiro-gh`, `kiro-local`, `kiro-notion`) or via the templating mechanism that generates them.
5. No behavioral change to `radio send` / `radio register` / `radio busy|ready|unregister` semantics other than fixing the cross-tab leak.

## Out of scope

- Refactoring the radio session-file format.
- Changing the SessionStart/SessionEnd/Stop hook wiring.
- The v0.2.1 release tag — PM will hold/reissue based on when this fix lands.

## Notes

- Symptom 2 has likely existed since the original v0.2.0 radio rollout.
- Symptom 1 is a **regression introduced by #99 / #95** (the `⏸️/▶️` emoji-rename feature) which merged into `main` on 2026-05-22 and is currently sitting in the unreleased v0.2.1 CHANGELOG block.

martin-conur:
## Summary

Fixes #102. Both `_rename_tab` (state-flip emoji prefix) and `cmd_send`'s wake-up now scope their `zellij action` side-effects to the role's own tab/pane id, not the focused tab.

- `_rename_tab` resolves the role's stable `tab_id` from its recorded `TAB=` name via `zellij action list-tabs --json` (parsed with `jq`, already a project dep) and renames via `rename-tab-by-id`. No more `zellij action rename-tab` that targets the currently focused tab.
- `cmd_send` resolves the recipient's `tab_id` the same way, then resolves a writable pane on it via `list-panes --json --tab` and writes `radio check\n` directly with `write-chars --pane-id`. No more `go-to-tab-name` (which stole the user's focus globally) and no more `write-chars` against the sender's pane context.
- Defense-in-depth: `_rename_tab` short-circuits when `$ZELLIJ_TAB` (the bare slug set by `task-work` / `task-pm` at pane creation) doesn't match the role's recorded tab — catches `TASK_FORCE_ROLE=other-role radio ready` invoked from the wrong pane even if the per-tab API call were to regress.

Body region synced across all 7 `*/bin/radio` copies; `tools/check-drift.sh` green.

## Test plan

- [x] New `tests/radio_cross_tab.bats` (4 tests) reproduces both symptoms in the issue:
  - bug 1: worker's `busy`/`ready` renames the worker's tab id only; PM's `TAB=` is untouched.
  - bug 2: `radio send --to pm` from worker B writes to pm's pane id (700), never worker-a's (800) or worker-b's (900), and never calls `go-to-tab-name`.
- [x] `tests/radio_tab_state.bats` (11 tests) updated to assert `rename-tab-by-id 7 <name>` instead of the old `rename-tab <name>`, plus two new tests for the `$ZELLIJ_TAB` safeguard.
- [x] `tests/radio_zellij.bats` (4 tests) updated to assert `write-chars --pane-id 700 radio check` and a no-`go-to-tab-name` refute; new test for "recipient's tab no longer exists" graceful skip.
- [x] `tests/radio_e2e.bats` updated to seed fixtures; still passes both bidirectional and cross-repo cases.
- [x] Full suite: 605/605 green.
- [x] `tools/check-drift.sh` green (16 groups).
- [ ] Live repro in a real zellij session (PM tab focused, worker auto-unregister; worker-a focused, worker-b → pm send) — to be verified during PR review.


martin-conur:
## Changes requested before merge

The two filed symptoms (#102) are fixed cleanly by `rename-tab-by-id` / `write-chars --pane-id` — that part of the design is right. But the **resolution path** still anchors on the visible tab name, which mutates on every idle↔busy flip via `_rename_tab`. That leaves three residual failure modes I'd like closed before merge:

1. **Stale TAB= race**. Between the sender reading the recipient's `TAB=⏸️ pm` and `cmd_send` calling `_zellij_tab_id_by_name`, the recipient may have transitioned idle→busy and rewritten its own session file. The lookup then misses, wake-up is silently queued. Message survives but the recipient never gets poked.
2. **Non-unique names**. `_zellij_tab_id_by_name` does `jq '... | .tab_id] | .[0]'` — if two tabs happen to share a name (e.g. the very state the current production bug just produced: two zellij tabs both displaying "▶️ pm"), the lookup grabs the first match and the wrong tab gets written to. The fix doesn't self-heal from an already-corrupted state.
3. **User-driven manual rename**. If the user renames a zellij tab themselves, the visible name diverges from the role's `TAB=` until the next state flip rewrites it. All sends to that role break in the meantime.

All three collapse if we **capture the stable `tab_id` once at registration** and reference it by id from then on. Display name becomes cosmetic.

## Requested changes

### 1. Resolve and persist `TAB_ID=` at register time
In `cmd_register`, after the session file is written:

- Resolve the tab's stable id once: query `zellij action list-tabs --json` for the tab whose `name` matches the `--tab` argument and pick its `tab_id`. (Reuse `_zellij_tab_id_by_name` — it's already there.)
- Persist as a new `TAB_ID=` line in the session file alongside `TAB=`.
- If the lookup fails (zellij unreachable, jq absent, tab not found), leave `TAB_ID=` empty and log it — don't fail registration. Callers must treat empty `TAB_ID=` as "fall back to name lookup" so non-zellij CI / first-register-before-rename paths still work.

### 2. Switch consumers to `TAB_ID=`
- `_rename_tab`: skip the `_zellij_tab_id_by_name "$current"` resolution — read `TAB_ID=` directly from the session file and pass it to `rename-tab-by-id`. Drop the now-redundant `TAB=` rewrite of the new name into the session file (it's purely cosmetic; nothing reads it functionally anymore). Keep painting `TAB=` so a human inspecting the session file can still see what name is visible, but stop treating it as load-bearing.
- `cmd_send` wake-up: read recipient's `TAB_ID=` from its session file, pass directly to `_zellij_pane_in_tab` (which is already id-based). Drop the by-name resolution. Keep the `TAB=` empty-check as a registered-vs-not gate.

### 3. Tests
- Add coverage for "recipient renamed mid-send" — sender wakes successfully even though `TAB=` in recipient's session file is one step behind. This is the central behavioural promise of the id-based lookup.
- Add coverage for "two tabs share a name" — sender still hits the right pane because we go through `TAB_ID=`, not the name. (Stub `list-tabs --json` to return two entries with identical `name` and different `tab_id`.)
- The existing `tests/radio_cross_tab.bats` cases should keep passing unchanged — id-based lookup is strictly stronger than name-based.

## Out of scope for this iteration

- Refactoring the session file format beyond adding `TAB_ID=`.
- Reacting to user-driven `rename-tab` of an already-registered role (we can detect drift and just accept it — name is cosmetic now).
- The defense-in-depth `$ZELLIJ_TAB` mismatch guard in `_rename_tab` — keep it as-is, it's still useful.

## Note on the current corrupted state

This PR, once landed, should self-heal the live corruption (two tabs displaying "▶️ pm") on next worker re-register, because `TAB_ID=` will pin each role to its actual tab regardless of display name collision. Worth verifying that in the live repro step.

Issue #107, opened 2026-05-22
task-done close-tab leaks to focused tab — closes PM tab when user switches away mid-cleanup


martin-conur:
## Problem

`task-done` runs `zellij action close-tab` (no target) as its final step. With no explicit tab id, `close-tab` targets the *currently focused* tab — same focused-tab class of bug we just closed out in `radio` (#102 / #103), in a different binary.

### Live repro (just observed)

1. Worker `worker-task-force-release-v021-amend` invoked `task-done --remove-worktree --force` to clean itself up after PR #105.
2. While `task-done` was running, the user switched focus to the **PM tab**.
3. `zellij action close-tab` fired from the worker's pane process, but at that moment the focused tab was PM's. **The PM tab got closed**; the worker's tab stayed open.

In the worst case, this destroys the user's active session including unsubmitted prompts / in-flight work. Same severity profile as #102.

### Code location

`*/bin/task-done` line ~133 (varies by loadout — claude-gh/claude-jira/claude-notion/kiro-gh/kiro-notion are at 133, claude-local/kiro-local at 153):

```bash
# Close current zellij tab
echo "Closing zellij tab..."
zellij action close-tab
```

Replicated across all 7 loadouts:
- `claude-gh/bin/task-done:133`
- `claude-local/bin/task-done:153`
- `claude-jira/bin/task-done:139`
- `claude-notion/bin/task-done:133`
- `kiro-gh/bin/task-done:133`
- `kiro-local/bin/task-done:153`
- `kiro-notion/bin/task-done:133`

### Why the #103 fix doesn't cover this

PR #103 scoped `radio`'s `rename-tab` and `write-chars` calls to per-id targeting, but `task-done` is a separate binary and was not touched. `task-done`'s close-tab still uses the global "focused tab" behavior — exact same root cause, separate code path.

## Solution

Drive close-tab by the worker's stable `TAB_ID=` (already persisted by `radio register` since #103), via `zellij action close-tab-by-id <id>` — which is a real zellij CLI command (confirmed with `zellij action --help | grep close-tab`).

### Concrete edits to `*/bin/task-done`

Replace the current block:

```bash
# Close current zellij tab
echo "Closing zellij tab..."
zellij action close-tab
```

With something like (worker can refine):

```bash
# Close this worker's zellij tab using the stable tab_id captured at
# `radio register` time (#103). Falling back to the old focused-tab
# `close-tab` would re-open the bug fixed there: if the user switches
[...]

PR #112, opened 2026-05-23
Radio: distinguish 'awaiting user input' from 'actively working' in tab state (#106)

### LINKED ISSUE #106: Radio: distinguish 'awaiting user input' from 'actively working' in tab state
## Problem

The radio idle/busy state model has exactly two buckets: `idle` (⏸️) and `busy` (▶️). The `busy` state is painted whenever Claude is *in turn* — which conflates two very different situations from the user's perspective:

1. **Actively working** — running tools, generating output, thinking. The user can step away; the worker will keep going.
2. **Waiting on the user** — permission prompt (`Allow this Bash command?`), `AskUserQuestion` dialog, `ExitPlanMode` approval. The worker is blocked until the user responds. No progress is being made.

Both currently display as `▶️ <slug>`. In a multi-worker setup, this means a user glancing at the zellij tab strip can't tell the difference between "PM is busy doing things" and "PM is sitting on a permission prompt that I need to click". The misreport erodes the at-a-glance signal that the `⏸️/▶️` feature (#95 / #99) was added to provide.

## Solution

Introduce a third state `awaiting` painted with a monochrome question-mark glyph (`❓︎` — black question mark + VS15 text-presentation selector to match the monochrome weight of ⏸️/▶️). Trigger it from Claude Code's `Notification` hook, which fires when Claude is prompting the user (permission gates, AskUserQuestion, ExitPlanMode).

> **Planning decision (emoji):** initial proposal was 🔔 (bell), but bell renders as a colored emoji on most fonts and breaks the monochrome style of ⏸️/▶️. Locked on `❓︎` (U+2753 + U+FE0E). If VS15 doesn't force text presentation on the user's terminal during smoke test, fall back to `⁇` (U+2047 double question mark) which is reliably monochrome across platforms.
>
> **Planning decision (Kiro):** Kiro's hook surface (`agentSpawn` / `userPromptSubmit` / `agentStop`) has no equivalent to Claude's `Notification` event. Kiro stays two-state for now. The radio body still gains the awaiting machinery (drift check requires byte-identical bodies across all 7 copies), but no `.kiro/hooks/radio-awaiting.json` is installed — nothing schedules `radio awaiting` from a Kiro session.

### State machine

```
              UserPromptSubmit
        ┌─────────────────────────┐
        │                         │
        ▼                         │
  ┌──────────┐               ┌─────────┐
  │  busy    │ Notification  │ awaiting│
  │ ▶️ slug  │──────────────▶│ ❓︎ slug │
  └──────────┘               └─────────┘
        │                         │
        │ Stop                    │ UserPromptSubmit (resume)
        ▼                         │ or Stop (cancelled)
  ┌──────────┐                    │
  │  idle    │◀───────────────────┘
  │ ⏸️ slug  │
  └──────────┘
```

Note on the awaiting → busy edge: Claude Code does not emit `UserPromptSubmit` when the user clicks "Allow" on a permission dialog, so `awaiting` may persist visually until `Stop` flips the tab to `⏸️`. This is acceptable — the post-action state is correct, the glyph just lingers a few extra seconds. Re-evaluate only if it proves noisy in practice.

### Wake-up policy (load-bearing — do not skip)

`cmd_send` currently gates wake-up on `STATE=idle`. The new `awaiting` state **must also be no-wake**, for the same reason `busy` is: writing `radio check\n` into a pane that's sitting at a permission prompt would have "r"/newline answer the prompt with the default (likely "no, ask again") or worse. The message still lands in the inbox; the recipient picks it up on their next manual `radio check` or next state flip back to `idle`.

Good news: the existing gate `[[ "$rstate" != "idle" ]]` (radio:415) **already covers awaiting correctly** — any non-idle state queues. So no behavioral change is needed in the gate itself. Only the log line is updated for clarity so awaiting and busy queues are distinguishable in the radio log.

#### 1. Radio script body — 7 copies, byte-identical (`# region:body`)

Files (drift-checked by `tools/check-drift.sh:27`):
`claude-{gh,jira,local,notion}/bin/radio`, `kiro-{gh,local,notion}/bin/radio`.

Edit one copy then sync the body region to the other six. Specific edits, by current line numbers in `claude-gh/bin/radio`:

- **`_strip_tab_prefix` (line 102–106)** — add `s="${s#❓︎ }"` to the strip list so repeated paints don't stack the glyph. (The original spec missed this; `_strip_tab_prefix` strips per-state.)
- **`_rename_tab` switch (line 235–239)** — add `awaiting) prefix="❓︎ " ;;` arm.
- **New `cmd_awaiting` (line ~352, next to `cmd_busy`)**:
  ```bash
  cmd_awaiting() { _silent_exit_if_no_role; _update_state awaiting; }
  ```
- **Subcommand dispatch (~line 538)** — add `awaiting) cmd_awaiting ;;`.
- **`cmd_send` log line (radio:415–418)** — keep the `!= "idle"` gate; update only the log message:
  ```bash
  if [[ "$rstate" != "idle" ]]; then
    local reason="busy"
    [[ "$rstate" == "awaiting" ]] && reason="awaiting input"
    _log "send: $to is $reason (state=$rstate) — queued (id=$id)"
    return 0
  fi
  ```
- **`usage()` text** — document `radio awaiting` alongside `radio busy` / `radio ready`. (Check whether this block is inside the `# region:body` sentinel; if yes, sync; if no, edit each file.)

Drift verification: `tools/check-drift.sh` must exit 0 after the edits.

#### 2. Claude task-init scripts — 4 copies

Files: `claude-{gh,jira,local,notion}/bin/task-init`.

In each `_radio_install_hooks` function, the existing jq merge declares 4 hooks (SessionStart / UserPromptSubmit / Stop / SessionEnd). Add a fifth:

```bash
local awaiting_cmd="radio awaiting"
```

then in the jq pipeline (alongside `--arg ups "$busy_cmd"` etc.):

```bash
--arg notif "$awaiting_cmd" \
'
…
| .hooks.Notification = add_radio(.hooks.Notification; $notif)
…
'
```

The `add_radio` jq helper is idempotent (entry added only if no existing hook command starts with `"radio "`), so re-running `task-init` on an upgraded project is safe.

#### 3. Host worktree `.claude/settings.json` (1 file)

> **Spec correction:** the original spec said "all 7 loadouts" for settings.json. There is no per-loadout `settings.json` — there is one `.claude/settings.json` in the active worktree, and the per-loadout `task-init` scripts merge hooks into target projects' `.claude/settings.json` at init time. The settings.json edit is therefore: one file in this worktree + the 4 claude-* task-init merges above.

Add a Notification block alongside the existing SessionStart / UserPromptSubmit / Stop / SessionEnd entries:

```json
"Notification": [
  {
    "hooks": [
      { "type": "command", "command": "radio awaiting" }
    ]
  }
]
```

#### 4. Kiro task-init scripts — leave untouched

`kiro-{gh,local,notion}/bin/task-init` are not modified. Kiro sessions never enter `awaiting`. Calling `radio awaiting` from a Kiro context still works (the subcommand exists in the synced body), but nothing in the Kiro install fires it.

#### 5. Tests

Extend the existing bats suite — no new file needed:

- **`tests/radio_tab_state.bats`** — add:
  - `cmd_awaiting writes STATE=awaiting and paints ❓︎ prefix`
  - `busy → awaiting → busy round-trip strips and repaints cleanly` (regression guard for the `_strip_tab_prefix` arm)
  - `awaiting → idle via cmd_ready paints ⏸️`
  - `awaiting → busy via cmd_busy paints ▶️`
  - `$ZELLIJ_TAB safeguard skips rename for awaiting on wrong tab`
- **`tests/radio_zellij.bats`** — add `cmd_send to awaiting recipient queues without write-chars` (mirror the existing "busy queues" test; assert log line contains `awaiting input`).
- **`tests/radio.bats`** — add `radio awaiting is silent no-op when TASK_FORCE_ROLE unset`.
- **`tests/radio_lifecycle.bats`** — (optional) `awaiting re-seeds session file when wiped`.

#### 6. README

In the "PM ↔ worker messaging (radio)" section (around lines 357–413):
- Document the three states (⏸️ idle / ▶️ busy / ❓︎ awaiting) and what triggers each.
- Note that awaiting recipients, like busy, queue messages silently and are picked up on next `radio check` / state flip back to idle.
- Add `radio awaiting` to the subcommand table.
- Add Notification to the lifecycle hook list.

## Files to Create/Modify

| File | Action | Purpose |
|------|--------|---------|
| `claude-{gh,jira,local,notion}/bin/radio` + `kiro-{gh,local,notion}/bin/radio` (7 copies, body-sync'd) | MODIFY | `_strip_tab_prefix`, `_rename_tab` switch, `cmd_awaiting`, dispatch, `cmd_send` log line, `usage()`. |
| `claude-{gh,jira,local,notion}/bin/task-init` (4 copies) | MODIFY | Add `Notification` → `radio awaiting` to the jq merge in `_radio_install_hooks`. |
| `.claude/settings.json` (host worktree) | MODIFY | Add Notification hook block. |
| `tests/radio_tab_state.bats` | MODIFY | State-machine + paint coverage for awaiting. |
| `tests/radio_zellij.bats` | MODIFY | Verify `cmd_send` queues to awaiting recipient without `write-chars`. |
| `tests/radio.bats` | MODIFY | Silent no-op coverage for `radio awaiting`. |
| `tests/radio_lifecycle.bats` | MODIFY (optional) | Re-seed parity. |
| `README.md` | MODIFY | Document three-state model and Notification hook. |

## Verification

1. **Drift:** `tools/check-drift.sh` exits 0 across the 7 radio bodies.
2. **Tests + shellcheck:**
   ```bash
   bats tests/
   shellcheck claude-gh/bin/radio claude-gh/bin/task-init
   ```
3. **Manual smoke (Claude loadout):**
   - Trigger a permission prompt in a worker session (e.g. `gh issue edit ...`). Verify tab repaints from `▶️ <slug>` to `❓︎ <slug>`. If the glyph renders colored, swap to `⁇` across the 4 edit sites and re-run drift.
   - Trigger `AskUserQuestion` (via `/planner`). Verify same paint flip.
   - Accept the prompt; verify tab flips to `▶️` while the turn continues and to `⏸️` once Stop fires.
   - From a second tab, `radio send --to <worker> --intent ping --body "test"` while worker is at `❓︎`. Verify the message file lands in `~/.task-force/radio/inboxes/<worker>/` but **no** `radio check` keystroke injection occurs, and the log says `awaiting input — queued`.
   - After approval, the next Stop should fire `radio check` and surface the queued message normally.
4. **CI:** existing bats / shellcheck / drift GitHub Actions all stay green.

## Acceptance criteria

1. Permission prompts and `AskUserQuestion` calls cause the role's tab to repaint with `❓︎` prefix; users see at a glance which sessions need their attention.
2. Resuming via `UserPromptSubmit` (e.g. answering an AskUserQuestion) transitions back to `▶️`. Stop transitions to `⏸️`. (Permission-prompt acceptance has no UserPromptSubmit — see note above.)
3. `radio send` to an `awaiting` recipient queues the message and does **not** fire `write-chars`; log line distinguishes the awaiting case from the busy case.
4. All existing bats / shellcheck / drift CI gates stay green.
5. Body region synced across all 7 `*/bin/radio` copies.

## Out of scope

- Wake-up payload changes — still `radio check\n`, still only fires for `idle` recipients.
- Cross-role notification fan-out.
- Persistence of the awaiting reason (which prompt is blocking).
- Distinguishing `AskUserQuestion` vs permission prompt vs `ExitPlanMode` — all three are "user needs to respond", all three deserve `❓︎`.
- Kiro loadouts — Kiro's hook surface has no Notification equivalent; revisit if/when it does.

## Notes

- Implementation risk is low — state machine is small, hook plumbing exists for `UserPromptSubmit` / `Stop`, wake-up gate needs no behavioral change. Main effort is bats coverage + body-region sync.
- Builds on #102 / #103 (TAB_ID-stable routing): tab rename happens via `rename-tab-by-id`, so the new `❓︎` prefix gets painted on the right tab regardless of focused tab.

---

**Planner notes (2026-05-22):** spec refined from the original draft based on codebase exploration:
- Confirmed there are 7 radio copies (verified byte-identical, md5 `64b072623c3f155049125d8d9f20c767`) and they're synced via `# region:body` markers checked by `tools/check-drift.sh:27`.
- Corrected the "7 settings.json" claim — there's 1 worktree `.claude/settings.json` plus 4 claude-* task-init jq merges (Kiro uses `.kiro/hooks/*.json` and is out of scope).
- Added the `_strip_tab_prefix` edit (the original spec missed it — prefix strip is per-glyph, will otherwise stack `❓︎ ⏸️ pm`).
- Locked emoji to `❓︎` (VS15) instead of 🔔; fallback `⁇` (U+2047) if VS15 doesn't force monochrome.

martin-conur:
Closes #106.

## Summary
- Adds a third radio state, `awaiting`, painted with a monochrome `❓︎` glyph (U+2753 + U+FE0E text-presentation selector). At-a-glance, users can now tell whether a worker is actively working (`▶️`) or blocked on a permission prompt / `AskUserQuestion` / plan approval (`❓︎`).
- Wired via Claude Code's `Notification` hook (`radio awaiting`); the existing `STATE != idle` send gate already covers awaiting correctly, so messages to an awaiting recipient queue silently — no stray keystrokes into a pending permission prompt.
- Kiro loadouts kept two-state (Kiro has no `Notification` analogue today).
- Drift-checked across all 7 `*/bin/radio` copies.

## Files changed
- 7 radio bodies (sync'd via `# region:body`): `_strip_tab_prefix` arm, `_rename_tab` switch case, `cmd_awaiting`, dispatch, `usage()`, distinct `cmd_send` log line for awaiting vs. busy.
- 4 claude-* `task-init` scripts: merge `Notification → radio awaiting` hook into target projects.
- 1 host `.claude/settings.json`: same Notification hook block.
- 4 bats files: state-machine paint, round-trip / no-stack regression guard, send-queues-without-write-chars + log-line distinction, silent no-op + STATE toggle, re-seed parity.
- `README.md`: new "Tab states" table + Notification hook documented.

## Test plan
- [x] `tools/check-drift.sh` exits 0 (7 radio bodies byte-identical).
- [x] Full `bats tests/` suite: 636/636 pass.
- [x] `shellcheck -x` (matching CI invocation): clean.
- [ ] Manual smoke (post-merge): trigger a permission prompt / `AskUserQuestion` in a worker; verify tab repaints to `❓︎ <slug>`. Verify `radio send` to that worker queues and logs `awaiting input` without `write-chars`. Verify `▶️` resumes after approval and `⏸️` on `Stop`.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -20,6 +20,16 @@
         ]
       }
     ],
+    "PreToolUse": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "radio busy"
+          }
+        ]
+      }
+    ],
     "Notification": [
       {
         "hooks": [
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -20,6 +20,16 @@
         ]
       }
     ],
+    "Notification": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "radio awaiting"
+          }
+        ]
+      }
+    ],
     "Stop": [
       {
         "hooks": [
--- a/README.md
+++ b/README.md
@@ -412,7 +412,8 @@ The `awaiting` glyph is driven by Claude Code's `Notification` hook, which fires
 |-------------------|-------------------------------|--------------------------------------------------------------|
 | `SessionStart`    | `radio register`              | Claims the role's session file for this tab                  |
 | `UserPromptSubmit`| `radio busy`                  | Marks the session busy while a turn is running               |
-| `Notification`    | `radio awaiting`              | Marks the session awaiting input on a permission/Ask prompt  |
+| `PreToolUse`      | `radio busy`                  | Flips back to busy when Claude resumes work after a permission "Allow" (no `UserPromptSubmit` fires in that case) |
+| `Notification`    | `radio awaiting`              | Marks the session awaiting input on a permission/Ask prompt (stdin payload filter skips idle-wait pings) |
 | `Stop`            | `radio ready && radio check`  | Marks idle and surfaces any queued messages                  |
 | `SessionEnd`      | `radio unregister`            | Cleans up the session file on real exit (intra-session events filtered) |
 
--- a/README.md
+++ b/README.md
@@ -385,24 +385,38 @@ Role names are addressable strings, not free-form: the PM is `pm`, and each work
 | `radio read <id>`                   | Print one message |
 | `radio ack <id>`                    | Mark it acknowledged so it stops showing up in `check` |
 | `radio register` / `radio unregister` | Add/remove this tab's session file (`~/.task-force/radio/sessions/<role>.info`) |
-| `radio ready` / `radio busy`        | Toggle this session's `STATE` field — drives the wake-up vs. queue decision on the sender side |
+| `radio ready` / `radio busy` / `radio awaiting` | Toggle this session's `STATE` field — drives the tab glyph (⏸️ / ▶️ / ❓︎) and the wake-up vs. queue decision on the sender side |
 | `radio orphans`                     | List session files whose heartbeat is >1h stale |
+
+### Tab states
+
+Each registered session paints a monochrome glyph on its zellij tab so you can tell at a glance what's happening:
+
+| Glyph | State      | Meaning                                                                              |
+|-------|------------|--------------------------------------------------------------------------------------|
+| ⏸️    | `idle`     | Turn ended, agent waiting for input or a queued message                              |
+| ▶️    | `busy`     | Agent is actively working — generating, running tools, thinking                       |
+| ❓︎    | `awaiting` | Agent is blocked on a user response (permission prompt, `AskUserQuestion`, plan mode) |
+
+The `awaiting` glyph is driven by Claude Code's `Notification` hook, which fires whenever Claude prompts the user. Kiro sessions have no equivalent event today and stay two-state.
 
 ### How wake-up works
 
-`radio send` reads the recipient's session file. If `STATE=idle`, it resolves the recipient's tab/pane id via `zellij action list-tabs --json` / `list-panes --json --tab` and writes `radio check\n` straight into that pane with `zellij action write-chars --pane-id` — no focus switch, so the sender's tab stays put. On the recipient's next turn end, the `Stop` hook (`radio ready && radio check`) surfaces the new message. If `STATE=busy`, the message is queued silently — no failed wake attempt, no interrupting the recipient mid-turn.
+`radio send` reads the recipient's session file. If `STATE=idle`, it resolves the recipient's tab/pane id via `zellij action list-tabs --json` / `list-panes --json --tab` and writes `radio check\n` straight into that pane with `zellij action write-chars --pane-id` — no focus switch, so the sender's tab stays put. On the recipient's next turn end, the `Stop` hook (`radio ready && radio check`) surfaces the new message. If `STATE=busy` or `STATE=awaiting`, the message is queued silently — no failed wake attempt, no risk of stray keystrokes answering a permission prompt, no interrupting the recipient mid-turn. The queued message is picked up on the recipient's next manual `radio check` or next flip back to `idle`.
 
 ### The hooks that make it work
 
 `task-init claude-*` writes these into your project's `.claude/settings.json` automatically:
 
-| Hook              | Command                       | Why                                            |
-|-------------------|-------------------------------|------------------------------------------------|
-| `SessionStart`    | `radio register`              | Claims the role's session file for this tab    |
-| `UserPromptSubmit`| `radio busy`                  | Marks the session busy while a turn is running |
-| `Stop`            | `radio ready && radio check`  | Marks idle and surfaces any queued messages    |
+| Hook              | Command                       | Why                                                          |
+|-------------------|-------------------------------|--------------------------------------------------------------|
+| `SessionStart`    | `radio register`              | Claims the role's session file for this tab                  |
+| `UserPromptSubmit`| `radio busy`                  | Marks the session busy while a turn is running               |
+| `Notification`    | `radio awaiting`              | Marks the session awaiting input on a permission/Ask prompt  |
+| `Stop`            | `radio ready && radio check`  | Marks idle and surfaces any queued messages                  |
+| `SessionEnd`      | `radio unregister`            | Cleans up the session file on real exit (intra-session events filtered) |
 
-For the kiro loadouts the same logic lives in `.kiro/hooks/` and runs off Kiro's equivalent triggers.
+For the kiro loadouts the same logic lives in `.kiro/hooks/` and runs off Kiro's equivalent triggers. Kiro has no `Notification` analogue today, so kiro sessions don't paint the `❓︎` glyph.
 
 ### Idle workers don't auto-act
 
--- a/claude-gh/bin/radio
+++ b/claude-gh/bin/radio
@@ -351,9 +351,35 @@ cmd_unregister() {
   _log "unregister role=$TASK_FORCE_ROLE"
 }
 
-cmd_ready()    { _silent_exit_if_no_role; _update_state idle; }
-cmd_busy()     { _silent_exit_if_no_role; _update_state busy; }
-cmd_awaiting() { _silent_exit_if_no_role; _update_state awaiting; }
+cmd_ready() { _silent_exit_if_no_role; _update_state idle; }
+cmd_busy()  { _silent_exit_if_no_role; _update_state busy; }
+
+# `radio awaiting` is wired to Claude Code's Notification hook, which fires for
+# both blocking prompts (permission gates, AskUserQuestion, ExitPlanMode) AND
+# idle-wait pings ("Claude is waiting for your input"). An unfiltered hook
+# would clobber STATE=idle with STATE=awaiting on every turn end, making the
+# ⏸️ glyph effectively unreachable in practice (#106 review). Read the JSON
+# payload piped on stdin and skip the state flip when the `message` field
+# indicates idle-wait, mirroring the cmd_unregister pattern (#107). Manual
+# CLI invocations (no piped payload) proceed unconditionally so the
+# subcommand remains usable from a shell.
+cmd_awaiting() {
+  _silent_exit_if_no_role
+  if [[ ! -t 0 ]]; then
+    local payload msg
+    payload=$(cat 2>/dev/null || true)
+    if [[ -n "$payload" ]] && command -v jq >/dev/null 2>&1; then
+      msg=$(printf '%s' "$payload" | jq -r '.message // empty' 2>/dev/null || true)
+      case "$msg" in
+        *waiting*your*input*|*waiting*for*input*|*idle*)
+          _log "awaiting: skipping (message=$msg — idle-wait, not blocking) role=$TASK_FORCE_ROLE"
+          return 0
+          ;;
+      esac
+    fi
+  fi
+  _update_state awaiting
+}
 
 cmd_send() {
   local to="" intent="" pr="" issue="" repo="" body=""
--- a/claude-gh/bin/radio
+++ b/claude-gh/bin/radio
@@ -19,6 +19,7 @@ Usage:
   radio register --role ROLE --tab TAB [--repo PATH] --agent claude|kiro [--loadout NAME]
   radio ready                            mark this role idle (and process pending)
   radio busy                             mark this role busy
+  radio awaiting                         mark this role awaiting user input
   radio unregister                       remove this role's session file
   radio orphans                          list session files with stale (>1h) heartbeat
 
@@ -103,6 +104,7 @@ _strip_tab_prefix() {
   local s="$1"
   s="${s#⏸️ }"
   s="${s#▶️ }"
+  s="${s#❓︎ }"
   printf '%s' "$s"
 }
 
@@ -233,9 +235,10 @@ _rename_tab() {
   fi
 
   case "$new_state" in
-    idle) prefix="⏸️ " ;;
-    busy) prefix="▶️ " ;;
-    *)    return 0 ;;
+    idle)     prefix="⏸️ " ;;
+    busy)     prefix="▶️ " ;;
+    awaiting) prefix="❓︎ " ;;
+    *)        return 0 ;;
   esac
   new_name="${prefix}${bare}"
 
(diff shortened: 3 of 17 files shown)


martin-conur:
## Changes requested before merge

The state-machine implementation is clean and the glyph choice is solid. Two issues need to be closed out before this is mergeable.

### 🔴 CRITICAL — Filter Notification hook by payload, not all-fires-paint-awaiting

Claude Code's \`Notification\` hook fires for multiple reasons — permission dialogs, \`AskUserQuestion\`, \`ExitPlanMode\`, **and idle-wait after a turn ends**. The PR's hook config is unfiltered (\`"command": "radio awaiting"\`), so the likely runtime sequence is:

1. Turn ends → \`Stop\` → \`radio ready\` → state=idle, paints \`⏸️\`
2. Idle-wait → \`Notification\` → \`radio awaiting\` → state=awaiting, paints \`❓︎\`

End state for every idle session: \`❓︎\`. The new state effectively *replaces* idle rather than augmenting it. The \`⏸️\` glyph becomes unreachable in practice.

The existing bats tests don't catch this because they exercise the radio CLI directly, never the hook chain.

**Required fix**: \`cmd_awaiting\` must read the JSON payload piped on stdin (same pattern \`cmd_unregister\` already uses for the \`clear\`/\`resume\` filter from #108) and only proceed when the \`message\` field indicates a *blocking* prompt. Suggested logic:

\`\`\`bash
cmd_awaiting() {
  _silent_exit_if_no_role
  # Notification hook fires for both blocking prompts AND idle-wait. We only
  # want to paint awaiting for the former. Read the stdin payload and skip
  # if the message looks like an idle-wait signal (any "waiting for ... input"
  # phrasing). Manual invocations (no stdin payload) proceed unconditionally.
  if [[ ! -t 0 ]]; then
    local payload msg
    payload=$(cat 2>/dev/null || true)
    if [[ -n "\$payload" ]] && command -v jq >/dev/null 2>&1; then
      msg=$(printf '%s' "\$payload" | jq -r '.message // empty' 2>/dev/null || true)
      case "\$msg" in
        *waiting*your*input*|*waiting*for*input*|*idle*)
          _log "awaiting: skipping (message=\$msg — idle-wait, not blocking) role=\$TASK_FORCE_ROLE"
          return 0
          ;;
      esac
    fi
  fi
  _update_state awaiting
}
\`\`\`

Mirror the pattern across all 7 \`*/bin/radio\` body copies; drift check must stay green.

**Required test coverage** (add to \`tests/radio_lifecycle.bats\` or equivalent):

- \`awaiting hook payload with message=\"Claude is waiting for your input\" → no state change\` — verify session file's STATE field is unchanged after the call.
- \`awaiting hook payload with message=\"Allow this Bash command?\" → state=awaiting\` — verify the blocking-prompt case still flips state.
- \`manual radio awaiting (no stdin) → state=awaiting\` — verify the hatch for direct CLI use stays open.

### 🟡 MEDIUM — \`awaiting → busy\` resume edge needs a hook

The spec correctly notes that Claude Code does NOT emit \`UserPromptSubmit\` when the user clicks "Allow" on a permission dialog. So the documented state machine edge:

> \`awaiting --UserPromptSubmit (resume)--> busy\`

…never fires in practice. Result: after "Allow", Claude resumes tool calls / generation, but the tab stays painted \`❓︎\` until \`Stop\` eventually flips it to \`⏸️\`. The user looking at the tab strip sees "needs my attention" while the worker is happily busy. That's a misreport in the *opposite direction* of the bug we set out to fix — arguably worse than the original.

**Required fix**: add a \`PreToolUse\` hook entry to \`.claude/settings.json\` (and the corresponding \`task-init\` jq merge) that fires \`radio busy\`. PreToolUse fires before each tool invocation, so as soon as Claude resumes work after "Allow", state flips back to busy. \`_update_state busy\` is idempotent when state is already busy, so the extra writes are harmless.

Update the README's "hooks that make it work" table to include the new PreToolUse row.

**Required test coverage**:

- \`PreToolUse hook fires radio busy → state=busy regardless of prior state\` — covers awaiting → busy and idle → busy.
- \`busy → busy via radio busy is a no-op for tab paint\` — defensive, confirms the idempotency.

### Out of scope for this iteration

- The VS15 fallback to \`⁇\` (planner's spec mentioned it as a fallback but the implementation locks in \`❓︎\` — accepting the cosmetic risk for now).
- Hook-integration tests at the full Claude-Code-hook-payload level (we'd need a richer test harness; CLI-level coverage of the payload filter is enough for now).

### Note on the broken-state risk

Until the CRITICAL fix lands, this PR would degrade the existing two-state UX (idle becomes invisible). Worth verifying live in a real Claude session after the fixes — fire a normal turn, confirm \`⏸️\` paints after Stop and does not get clobbered by Notification.

martin-conur:
Both review issues addressed in 4031e9f.

### 🔴 CRITICAL — Notification payload filter

`cmd_awaiting` now reads the JSON payload piped on stdin and skips the state flip on idle-wait messages (mirroring the `cmd_unregister` pattern from #107). The filter checks the `message` field for `*waiting*your*input*`, `*waiting*for*input*`, or `*idle*`. Manual CLI invocations (no stdin) proceed unconditionally so the subcommand stays usable from a shell. Synced across all 7 `*/bin/radio` copies — drift check clean.

7 new bats cases in `radio_lifecycle.bats`:
- `awaiting skips when stdin payload has message='Claude is waiting for your input' (idle-wait)`
- `awaiting skips when stdin payload has message='Claude is waiting for user input'`
- `awaiting proceeds when stdin payload has a blocking-prompt message (permission gate)`
- `awaiting proceeds when stdin payload has a blocking-prompt message (AskUserQuestion)`
- `awaiting proceeds when stdin is empty (manual CLI invocation)`
- `awaiting proceeds when stdin payload is not valid JSON`
- `awaiting proceeds when payload JSON has no message field`

### 🟡 MEDIUM — `PreToolUse → radio busy` for awaiting → busy resume

Added a `PreToolUse` hook firing `radio busy` to:
- Host worktree `.claude/settings.json`
- All 4 claude-* `task-init` jq merges (reuses the existing `$busy_cmd` arg)

Next tool invocation after a permission "Allow" now flips the tab back to ▶️.

2 new bats cases in `radio_tab_state.bats`:
- `busy → busy is a no-op for STATE and paints the same play prefix` (idempotency)
- `awaiting → busy via PreToolUse-style call flips STATE and paints ▶️` (resume path)

README hooks table updated to include the new PreToolUse row + a note on the Notification filter.

### Verification
- ✅ `tools/check-drift.sh` (16 groups checked)
- ✅ Full `bats tests/` (645 passing, +9 new)
- ✅ `shellcheck -x` (CI invocation) — clean

PR #118, opened 2026-05-24
Fix #117: preserve TAB_ID across re-register; persist TAB_ID in $INFO_FILE

### LINKED ISSUE #117: task-done tab leak: TAB_ID gets clobbered when radio re-registers on session re-init (--auto / non-auto regression, broader than #109)
## Problem

`task-done --remove-worktree` leaves the worker's zellij tab orphaned. Observable symptom matches #109 (`task-done` prints `"Skipping zellij close-tab (no tab id captured...)"` and the tab survives), but the **root cause is different** from #109's "40+-unregister-loop" and different from this issue's original (now-corrected) `SessionEnd`-timing hypothesis.

### Live evidence — radio log timeline (2026-05-24)

```
17:46:43  register role=worker-task-force-probe-awaiting-hooks tab=probe-awaiting-hooks tab_id=12   ← good register
…         (worker runs, radio busy paints tab to `▶️ probe-awaiting-hooks`)
18:02:22  register: could not resolve tab_id for tab=probe-awaiting-hooks
          (zellij list-tabs returned no match)
18:02:22  register role=worker-task-force-probe-awaiting-hooks tab=probe-awaiting-hooks tab_id=    ← TAB_ID clobbered to empty
18:02:22  _rename_tab: no tab_id for role worker-task-force-probe-awaiting-hooks — skipping
…         (session continues; every subsequent _rename_tab / cmd_send logs "no TAB_ID")
19:43:31  send: worker-task-force-probe-awaiting-hooks has no TAB_ID — queued
19:58:55  unregister role=worker-task-force-probe-awaiting-hooks                                   ← task-done runs
          (TAB_ID was already empty in session file → close-tab-by-id skipped → tab leaks)
```

### Root cause

At 18:02:22 the worker's Claude session re-initialized (`/compact`, `/clear`, `/resume`, or a fresh `claude` launched in the same tab — any event that fires `SessionStart` a second time). The `SessionStart` hook ran `radio register --tab $ZELLIJ_TAB`, where `$ZELLIJ_TAB` still resolved to the **original** tab name `probe-awaiting-hooks` (the env var doesn't update when the tab is renamed). Inside `cmd_register`, the `zellij list-tabs` lookup tried to match `probe-awaiting-hooks` literally — but the tab had since been renamed by `radio busy` paint to `▶️ probe-awaiting-hooks`. The name lookup failed; `cmd_register` fell into the "could not resolve" path and **overwrote `TAB_ID=12` with `TAB_ID=` (empty)** in the session file (`claude-gh/bin/radio:280-301`).

From that point on:
- Every `_rename_tab` call skipped silently (no busy/idle paint updates for the rest of the session).
- Every `cmd_send` to this worker queued instead of waking it (`send: … has no TAB_ID — queued`).
- `task-done` later read the now-empty `TAB_ID` from the session file → `_WORKER_TAB_ID=""` → "Skipping zellij close-tab" → tab leaks.

### Affects both --auto and non-auto workers

The original issue body blamed `SessionEnd` timing in `--auto` mode. **That diagnosis was wrong.** Live reproduction in a non-`--auto` session shows the same leak — because any Claude session that re-initializes at least once (`/compact`, `/clear`, `/resume`, a fresh `claude` in the same tab) will fire `SessionStart` again and clobber `TAB_ID`. `/compact` in particular is a routine action; basically any non-trivial session hits this.

### Why #108/#109 don't cover this

PR #108 (TAB_ID-stable `close-tab-by-id` in `task-done`) is right and shouldn't be widened — its strict-skip is what prevents #107 from re-opening. PR #109 (`cmd_unregister` skip on `clear`/`resume` reasons) targeted the symmetric problem on the *unregister* side. The *register* side has an analogous-but-distinct failure mode that's never been fixed.

## Proposed fix

Two layers, ideally one PR (close fits):

### Layer 1 — stop `cmd_register` from clobbering an existing valid `TAB_ID`

In `claude-gh/bin/radio:cmd_register` (~line 280–301), when the `zellij list-tabs` lookup fails:
- If a session file already exists with a non-empty `TAB_ID=`, **preserve it** in the rewrite.
- Only write `TAB_ID=` (empty) on the *first* register (no prior session file).

Pseudocode for the change:

```bash
# Inside cmd_register, after zellij lookup:
if [[ -z "$tab_id" ]]; then
  # Lookup failed. Preserve existing TAB_ID if present.
  local existing_tab_id=""
  [[ -f "$f" ]] && existing_tab_id=$(awk -F= '/^TAB_ID=/ { print $2; exit }' "$f")
  if [[ -n "$existing_tab_id" ]]; then
    tab_id="$existing_tab_id"
    _log "register: zellij list-tabs miss for tab=$tab — preserving existing TAB_ID=$tab_id role=$TASK_FORCE_ROLE"
  else
    _log "register: could not resolve tab_id for tab=$tab (zellij list-tabs returned no match)"
  fi
fi
```

(Optional alternative: strip the radio paint prefix off `$tab` before the lookup — `tab_for_lookup=$(echo "$tab" | sed -E 's/^([⏸▶❓][^ ]* )?//')`. Less robust to future paint additions.)

### Layer 2 — defense-in-depth: persist `TAB_ID` in `$INFO_FILE`

`task-work` already knows the tab id when it creates the tab (prints it: `"12"` after `Opening zellij tab '<slug>'...`). Persist it in `$INFO_FILE` so `task-done` has an authoritative source independent of the radio session file's mid-life state:

```bash
# task-work:208 — extend the printf
printf 'BASE_BRANCH=%s\nSLUG=%s\nGH_URL=%s\nTAB_ID=%s\n' \
  "$BASE_BRANCH" "$SLUG" "$GH_URL" "$TAB_ID" > "$INFO_FILE"
```

```bash
# task-done:107-113 — try $INFO_FILE first
_WORKER_TAB_ID=""
if [[ -f "$INFO_FILE" ]]; then
  _WORKER_TAB_ID=$(awk -F= '/^TAB_ID=/ { print $2; exit }' "$INFO_FILE")
fi
# Fall back to radio session file for migrated workers / unusual states
if [[ -z "$_WORKER_TAB_ID" && -n "${ZELLIJ:-}" && -n "${TASK_FORCE_ROLE:-}" ]]; then
  _SESSION_FILE="${TASK_FORCE_HOME:-$HOME/.task-force}/radio/sessions/${TASK_FORCE_ROLE}.info"
  if [[ -f "$_SESSION_FILE" ]]; then
    _WORKER_TAB_ID=$(awk -F= '/^TAB_ID=/ { print $2; exit }' "$_SESSION_FILE")
  fi
fi
```

Layer 1 alone fixes the user-facing `task-done` symptom AND the parallel session-degradation symptoms (no mid-session repaints, queued sends). Layer 2 makes the system robust to any future radio-session-file accident — `$INFO_FILE` is owned by `task-work`/`task-done` and is never touched during a session.

## Files to Create/Modify

| File | Action |
|------|--------|
| `claude-{gh,jira,local,notion}/bin/radio` + `kiro-{gh,local,notion}/bin/radio` (7 copies, body-sync'd) | MODIFY — Layer 1 in `cmd_register` |
| `claude-{gh,jira,local,notion}/bin/task-work` (4 copies) | MODIFY — Layer 2 (`TAB_ID=` into `$INFO_FILE`) |
| `kiro-{gh,local,notion}/bin/task-work` (3 copies) | MODIFY — same |
| `claude-{gh,jira,local,notion}/bin/task-done` (4 copies) | MODIFY — Layer 2 (read from `$INFO_FILE` first) |
| `kiro-{gh,local,notion}/bin/task-done` (3 copies) | MODIFY — same |
| `tests/radio_lifecycle.bats` (or new) | ADD — Layer 1 coverage: re-register preserves existing TAB_ID when lookup fails |
| `tests/task_done_close_tab.bats` (or extend existing) | ADD — Layer 2 coverage |
| `tools/check-drift.sh` | RUN — verify body regions stay synced |

## Acceptance criteria

1. **Layer 1**: When `cmd_register` runs a second time in the same tab (simulated by renaming the tab to include a paint prefix and re-invoking `radio register`), the existing `TAB_ID` is preserved, not clobbered to empty. Log line distinguishes "preserved" from the original first-time-no-match case.
2. **Layer 2**: `task-work` writes `TAB_ID=<id>` into `$INFO_FILE` at worktree creation time.
3. **Layer 2**: `task-done` reads `TAB_ID` from `$INFO_FILE` first; falls back to radio session file; falls back to silent-skip.
4. **Regression scenario passes**: a worker that has had at least one `/compact` (or other session re-init) during its life still has its tab closed correctly by `task-done --remove-worktree`.
5. Original #107/#108 scenarios still pass (worker tab focused → close worker tab; PM tab focused mid-cleanup → still close worker tab, not PM).
6. **Mid-session side-effects also fixed**: after `/compact`, `_rename_tab` still paints state correctly, and `cmd_send` still wakes the recipient instead of falling through to the "no TAB_ID — queued" branch.
7. Loadout drift check green across all 7 `radio` + 7 `task-work` + 7 `task-done` copies.
8. ShellCheck green; full bats suite green.

## Out of scope

- Reworking radio's broader lifecycle semantics.
- Re-opening #109 — distinct failure mode.
- Migrating pre-existing live workers (none in flight at file time).

martin-conur:
Fixes #117.

## Summary

`task-done --remove-worktree` was leaving worker zellij tabs orphaned whenever the worker's Claude session had re-initialized at least once (`/compact`, `/clear`, `/resume`, or a fresh `claude` in the same tab). Live log evidence in #117 traced the regression to `cmd_register` blindly rewriting `TAB_ID=` to empty when the post-paint tab name no longer matched the bare slug `$ZELLIJ_TAB` passes in.

Two layers:

- **Layer 1** — `cmd_register` (`radio`): on a zellij `list-tabs` miss, preserve an existing non-empty `TAB_ID=` from the session file instead of clobbering it. The "could not resolve" log line stays for the first-register-no-match case; a distinct "preserving existing TAB_ID=N" line covers the re-register path. Stops the cascade that breaks `_rename_tab` repaints and `cmd_send` wake-ups for the rest of the worker's life.
- **Layer 2** — `task-work` + `task-done`: persist the worker's `TAB_ID=` into `$INFO_FILE` after `aw_launch_tab` (via a new `aw_zellij_tab_id_by_name` helper in `lib/zellij-tab.sh`). `task-done` reads `$INFO_FILE` first, falls back to the radio session file for pre-fix workers. `$INFO_FILE` is owned by `task-work`/`task-done` and is never touched by radio, so close-tab survives any future mid-life radio-session-file accident.

Synced across all 7 `radio` copies, all 7 `task-work` copies, and all 7 `task-done` copies; drift check manifest extended with the new `task-work|info-tab-id` group.

## Test plan

- [x] `tests/radio_lifecycle.bats` — three new tests cover Layer 1: re-register preserves TAB_ID on lookup miss; first-time miss still logs "could not resolve"; re-register with empty existing TAB_ID falls through (no "preserving" log).
- [x] `tests/task_done_close_tab.bats` — three new tests cover Layer 2: prefer `$INFO_FILE`; fall back to session file when `$INFO_FILE` has no TAB_ID; close still works when the session file is missing.
- [x] `tests/claude_gh_task_work.bats` — two new tests cover the `$INFO_FILE` TAB_ID write (with and without a zellij fixture).
- [x] `tools/check-drift.sh` — 17/17 groups green (new `info-tab-id` group across all 7 task-work loadouts).
- [x] Full bats suite — 636/636 passing.
- [x] All four existing `#107` close-tab scenarios still pass.

### DIFF
--- a/claude-gh/bin/radio
+++ b/claude-gh/bin/radio
@@ -293,11 +293,33 @@ cmd_register() {
   local tab_id=""
   if [[ -n "${ZELLIJ:-}" ]] && command -v zellij >/dev/null 2>&1; then
     tab_id=$(_zellij_tab_id_by_name "$tab" || true)
-    [[ -z "$tab_id" ]] && _log "register: could not resolve tab_id for tab=$tab (zellij list-tabs returned no match)"
   fi
 
   local f
   f=$(_session_file "$role")
+
+  # Re-register protection (#117): `SessionStart` fires again on `/compact`,
+  # `/clear`, `/resume`, or a fresh `claude` in the same tab — each time
+  # passing the original $ZELLIJ_TAB (the bare slug captured at pane creation).
+  # By then `_rename_tab` has painted the visible name to "⏸️ <slug>" or
+  # "▶️ <slug>", so a literal-name lookup misses and `tab_id` is empty. If we
+  # blindly rewrote the session file we'd clobber a valid TAB_ID with "",
+  # killing all subsequent _rename_tab repaints and cmd_send wake-ups for the
+  # rest of the worker's life — and finally leaking the tab when task-done
+  # reads the now-empty TAB_ID (#117). Preserve the existing TAB_ID instead.
+  if [[ -z "$tab_id" ]]; then
+    local existing_tab_id=""
+    if [[ -f "$f" ]]; then
+      existing_tab_id=$(_session_field "$f" TAB_ID || true)
+    fi
+    if [[ -n "$existing_tab_id" ]]; then
+      tab_id="$existing_tab_id"
+      _log "register: zellij list-tabs miss for tab=$tab — preserving existing TAB_ID=$tab_id role=$role"
+    else
+      _log "register: could not resolve tab_id for tab=$tab (zellij list-tabs returned no match)"
+    fi
+  fi
+
   printf 'ROLE=%s\nTAB=%s\nTAB_ID=%s\nREPO=%s\nSTATE=idle\nLAST_HEARTBEAT=%s\nAGENT=%s\nLOADOUT=%s\n' \
     "$role" "$tab" "$tab_id" "$repo" "$(_now)" "$agent" "$loadout" \
     | _atomic_write "$f"
--- a/claude-gh/bin/task-done
+++ b/claude-gh/bin/task-done
@@ -104,8 +104,18 @@ fi
 # focused at call time — same focused-tab bug class as #102/#103 in a
 # different binary; if the user switches focus mid-cleanup, the wrong tab
 # (e.g., PM's) gets closed and an active session can be destroyed (#107).
+#
+# Preferred source: $INFO_FILE, which task-work writes at tab creation and
+# nobody else touches. Independent of the radio session file's mid-life
+# state, so an in-session SessionStart re-register that clobbered the
+# session file's TAB_ID= (#117) can't break cleanup. Fall back to the
+# radio session file for workers whose task-work pre-dates the $INFO_FILE
+# TAB_ID write.
 _WORKER_TAB_ID=""
-if [[ -n "${ZELLIJ:-}" && -n "${TASK_FORCE_ROLE:-}" ]]; then
+if [[ -n "${ZELLIJ:-}" && -f "$INFO_FILE" ]]; then
+  _WORKER_TAB_ID=$(awk -F= '/^TAB_ID=/ { print $2; exit }' "$INFO_FILE")
+fi
+if [[ -z "$_WORKER_TAB_ID" && -n "${ZELLIJ:-}" && -n "${TASK_FORCE_ROLE:-}" ]]; then
   _SESSION_FILE="${TASK_FORCE_HOME:-$HOME/.task-force}/radio/sessions/${TASK_FORCE_ROLE}.info"
   if [[ -f "$_SESSION_FILE" ]]; then
     _WORKER_TAB_ID=$(awk -F= '/^TAB_ID=/ { print $2; exit }' "$_SESSION_FILE")
--- a/claude-gh/bin/task-work
+++ b/claude-gh/bin/task-work
@@ -238,6 +238,11 @@ fi
 # owned by task-work / task-done and is never written by radio, so it stays
 # authoritative regardless of any SessionStart/SessionEnd re-register churn
 # inside the worker.
+#
+# Race-safe: aw_zellij_tab_id_by_name strips ⏸/▶/❓ paint prefixes off the
+# zellij-side .name before matching against $SLUG, so a `radio busy` paint
+# that fires between aw_launch_tab returning and this lookup (worker's
+# SessionStart → first prompt) can't make the match miss (#117 review).
 INFO_TAB_ID=$(aw_zellij_tab_id_by_name "$SLUG" || true)
 if [[ -n "$INFO_TAB_ID" ]]; then
   printf 'TAB_ID=%s\n' "$INFO_TAB_ID" >> "$INFO_FILE"
--- a/claude-gh/bin/task-work
+++ b/claude-gh/bin/task-work
@@ -230,3 +230,16 @@ else
   echo "Started worker in $WORKTREE_DIR (branch: $BRANCH, base: $BASE_BRANCH)"
 fi
 # endregion:worktree-creation-post-info
+
+# region:info-tab-id
+# Persist the zellij tab_id of the worker's tab into $INFO_FILE so task-done
+# can scope `zellij action close-tab-by-id` to *this* worker even if the
+# radio session file's TAB_ID gets clobbered mid-life (#117). $INFO_FILE is
+# owned by task-work / task-done and is never written by radio, so it stays
+# authoritative regardless of any SessionStart/SessionEnd re-register churn
+# inside the worker.
+INFO_TAB_ID=$(aw_zellij_tab_id_by_name "$SLUG" || true)
+if [[ -n "$INFO_TAB_ID" ]]; then
+  printf 'TAB_ID=%s\n' "$INFO_TAB_ID" >> "$INFO_FILE"
+fi
+# endregion:info-tab-id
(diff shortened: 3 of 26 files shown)


martin-conur:
## Review — one nit, otherwise LGTM

Code is clean, test coverage hits the actual #117 regression scenario, drift synced. One non-blocking nit worth tightening before merge:

### Race in `task-work`'s post-launch TAB_ID lookup

`aw_zellij_tab_id_by_name "\$SLUG"` runs right after `aw_launch_tab`. If Claude's SessionStart-hook → `radio register` → first prompt → `radio busy` chain repaints the tab to `"▶️ <slug>"` before `task-work`'s lookup actually executes, the literal-name match misses and `TAB_ID` doesn't make it into `\$INFO_FILE`. We then degrade silently to the radio session file fallback — which is exactly the unreliable source Layer 2 was supposed to insulate us from.

Worst case isn't a regression (we're no worse off than before this PR), but it does undermine the "INFO_FILE is authoritative" guarantee Layer 2 was added to provide. Worth closing.

### Suggested fixes (worker's call on which is cleanest)

1. **Capture tab_id from `aw_launch_tab` directly.** Zellij prints the tab_id on creation (we already see this in `task-work` output — e.g., the literal `"12"` line after `"Opening zellij tab '<slug>'..."`). If `aw_launch_tab` can return that id (set a variable, echo it, etc.), `task-work` writes it to `$INFO_FILE` with no name lookup involved. Most robust — eliminates the race entirely.
2. **Strip paint prefixes in `aw_zellij_tab_id_by_name`'s jq filter.** Match on `.name | sub("^[⏸▶❓][^ ]* "; "") == $n`. Survives any paint state at lookup time. Brittle if a future glyph is added without updating the strip set.
3. **Short retry loop.** `for _ in 1 2 3; do tab_id=$(aw_zellij_tab_id_by_name ...); [[ -n "$tab_id" ]] && break; sleep 0.05; done`. Cheap, race-tolerant, but doesn't help if the paint actually wins the race — which is the exact scenario we care about.

Option 1 is the cleanest IMO — closes the race rather than working around it, and the tab_id is right there in zellij's own output. Option 2 is the fallback if option 1 isn't feasible without restructuring `aw_launch_tab`.

Either way:
- Add a test that asserts `TAB_ID=` lands in `$INFO_FILE` even when the tab name has been pre-painted at the time of the lookup (use the same `STUB_ZELLIJ_TABS_JSON='[{"name":"▶️ <slug>",...}]'` trick from `tests/radio_lifecycle.bats`).
- Update the comment in `task-work`'s `# region:info-tab-id` block to note this is race-safe.

No other changes needed. Once this lands, approve + merge as soon as CI is green.

Issue #128, opened 2026-05-25
radio: opt-in auto-submit on wake-up for --auto workers (CR not LF)


martin-conur:
## Problem

`radio send` wakes an idle recipient by writing `\"radio check\"$'\n'` into its zellij pane via `zellij action write-chars` (see `<loadout>/bin/radio` cmd_send, after the send-gate). Verified empirically (2026-05-25 with `stty raw -echo; cat -v` against `write-chars --pane-id`): `$'\n'` is delivered as raw LF (`0x0A`), `$'\r'` as raw CR (`0x0D`). Raw-mode TUIs like Claude Code's input box bind **Enter to CR**, not LF — so today's wake-up leaves `radio check` sitting in the prompt and the user has to press Enter for the recipient to actually process the radio message.

For the **PM tab**, that human gate is desirable: PM is often mid-task, and surfacing radio's `radio check` into a partially-typed prompt would corrupt PM's work-in-progress. Keep it manual.

For **workers**, the gate is friction with no upside: workers spend most of their life idle, waiting on PM to ping back (`changes-requested` / `approved-and-merged`). Nothing else is queued in the pane. Auto-submitting on a `--auto`-launched worker is safe and consistent with the existing autonomy signal.

## Solution

Opt-in auto-submit, gated on the autonomy signal already present (`task-work --auto`).

1. **`task-work` (all 7 loadouts, region `radio-env-injection`)**: when `\$AUTO_MODE` is set, append `TASK_FORCE_AUTO_SUBMIT=1` to `RADIO_ENV_PREFIX`. `task-pm` never sets this, so PM stays human-gated by default.
2. **`radio register` (all 7, `cmd_register`)**: if `\$TASK_FORCE_AUTO_SUBMIT=1` is present in the env, write `AUTO_SUBMIT=1` as an extra line in the session `.info` file alongside `STATE`, `TAB_ID`, etc. Mirror the same in `_ensure_session_file` so self-heal preserves the flag.
3. **`radio send` (all 7, `cmd_send`)**: read `AUTO_SUBMIT` from the recipient's session file via `_session_field`. If `=1`, write `\"radio check\"$'\r'` (CR — triggers submit). Otherwise keep `\"radio check\"$'\n'` (LF — visible, awaits Enter).

The send-gate (STATE != idle → queue silently) is unchanged. Auto-submit only changes the byte used by an already-running wake-up.

## Files to Create/Modify

| File | Action | Purpose |
|------|--------|---------|
[...]

PR #129, opened 2026-05-25
radio: opt-in auto-submit on wake-up for --auto workers (CR not LF) (#128)

martin-conur:
## Summary

Implements #128. `radio send`'s zellij `write-chars` wake-up writes `radio check` + a terminator into the recipient's pane. The current LF (`\n`) leaves it sitting in Claude Code's input box until the user presses Enter — desirable for PM (protects a partially-typed prompt) but pure friction for workers, which idle most of their life waiting on PM pings.

This PR adds an **opt-in** auto-submit gated on the existing autonomy signal:

- `task-work --auto` (all 7 loadouts, `radio-env-injection` region): exports `TASK_FORCE_AUTO_SUBMIT=1` alongside the existing radio env vars.
- `radio register` + `_ensure_session_file` (all 7): if `$TASK_FORCE_AUTO_SUBMIT=1` is in env, write `AUTO_SUBMIT=1` into the session `.info` file; self-heal preserves it.
- `radio send`: reads `AUTO_SUBMIT` from the recipient's session file. `=1` → CR (`\r`, triggers Enter). Anything else (including legacy / absent) → LF (current behaviour).

Send-gate (STATE != idle → queue silently) is unchanged. PM never sets the env var, so its wake-ups stay LF-gated.

## Test plan

- [x] `./run_tests.sh` — all 651 tests pass, including new `tests/radio_auto_submit.bats` (8 tests pinning the contract).
- [x] `tools/check-drift.sh` — all 17 region groups clean (radio body + task-work radio-env-injection edits are byte-identical across all 7 loadouts).
- [ ] Manual: launch `task-work … --auto` worker, verify `~/.task-force/radio/sessions/worker-*.info` contains `AUTO_SUBMIT=1`, then `radio send` from PM and confirm the worker processes the message without a manual Enter.
- [ ] Manual negative: launch a default (non-`--auto`) worker, confirm no `AUTO_SUBMIT` line and that `radio check` still awaits Enter.
- [ ] Manual: confirm PM never auto-submits, even when pinged by an `--auto` worker.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Added
+
+- **Opt-in auto-submit on radio wake-up for `--auto` workers.** `radio send`'s zellij `write-chars` wake-up has always written `radio check\n` (LF) into the recipient's pane, leaving it sitting in the input box until the user pressed Enter. For workers launched via `task-work --auto` — already an explicit autonomy opt-in, and idle most of their life waiting on PM — that human gate is friction with no upside. `task-work --auto` now exports `TASK_FORCE_AUTO_SUBMIT=1`; `radio register` / `_ensure_session_file` persist that as `AUTO_SUBMIT=1` in the session file; `cmd_send` reads it and switches the wake-up byte to CR (`\r`), which Claude Code's raw-mode input binds to Enter. PM and default (non-`--auto`) workers leave the flag unset and keep the LF wake-up, so a `radio check` ping can never corrupt a partially-typed prompt. (#128)
+
 ## [0.2.2] — 2026-05-23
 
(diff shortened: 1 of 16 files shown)


martin-conur:
LGTM — implementation matches the spec, design is tight, tests pin the right contract.

Two minor observations for the record (not blockers):

1. **AUTO_SUBMIT is per-launch, not per-worktree.** If the user kills claude and restarts with bare `claude` in the same worktree, the next `radio register` drops the flag because `TASK_FORCE_AUTO_SUBMIT` is no longer in the env. Consistent with how `--auto` is currently scoped, but if anyone hits this in practice we could persist the flag in `$INFO_FILE` and have `radio register` honor it.

2. **Partial-input edge under AUTO_SUBMIT=1.** If a worker has partially-typed text in its input box when PM pings, the CR wake-up will submit `<typed text>radio check` together. The STATE=idle send-gate makes this unlikely (workers are idle precisely when nothing's mid-turn), and it's consistent with the autonomy contract — but worth keeping in mind if anyone ever reports a "weird prompt got submitted" issue against `--auto` workers.

Merging.

PR #132, opened 2026-05-26
radio: read should auto-ack (move inbox→processed); add --peek escape hatch (#131)

### LINKED ISSUE #131: radio: `read` should auto-ack (move inbox→processed); add --peek escape hatch
## Problem

The current `radio read <id>` / `radio ack <id>` split is a workflow footgun. `read` prints the message body; `ack` moves the file from `inbox/` to `processed/`. The two are intended to be used together — there is no real workflow that legitimately wants to read a message and *not* mark it acknowledged.

The failure pattern, observed in production:

1. Worker handles a `changes-requested` ping (pushes a fix, sends `re-review-requested` back to PM).
2. Worker forgets to run `radio ack <id>` afterwards. The original message stays in `inbox/`.
3. Worker's `Stop` hook auto-runs `radio check` on every idle turn. Each subsequent check re-lists the stale message identically.
4. Worker reads the same line as "still only the original changes-requested" and concludes "no new signal" — but in fact the line is the *old* one and the file's mtime hasn't moved. PM's silence is indistinguishable from PM not having pinged at all.
5. The user has to step in manually ("check it man i dont see you doing it") to break the loop.

The structural problem is that the inbox state machine has three legitimate states (`unread`, `read-but-not-acted`, `acted`) but only two storage buckets (`inbox/`, `processed/`), and the transition between the middle state and the third is a manual command that nothing forces you to run.

## Solution

Collapse `read` and `ack` into a single beat by default. Add an explicit `--peek` (alias: `--no-ack`) escape hatch for the rare "I just want to confirm what's in there without consuming it" case.

- **`radio read <id>` (default):** print the message body to stdout AND `mv` the file from `inbox/` to `processed/`. Same final state as today's `radio read <id> && radio ack <id>` pairing.
- **`radio read --peek <id>` / `radio read --no-ack <id>`:** print the body, do NOT move the file. Read-only operation. For inspection / debugging / "did PM ping me 3 days ago, what did they say."
- **`radio ack <id>` (legacy command):** keep working. Stays idempotent: if the message is already in `processed/` because a prior `read` already moved it, `ack` exits 0 with a status note ("already processed") rather than failing. This protects existing scripts / muscle memory that pair `read` + `ack`.

The user's worker-side handling discipline (saved as memory `feedback-radio-ack-after-handling`) becomes "run `radio read <id>` as part of handling" — which is the natural beat anyway, since you need to see the body to act on it. The system now enforces the rest.

## Files to Create/Modify

| File | Action | Purpose |
|------|--------|---------|
| 7× `<loadout>/bin/radio` (region:`body`, `cmd_read`) | MODIFY | Parse `--peek` / `--no-ack` from argv. Print body as today. If neither escape flag was passed AND the message was in `inbox/` (not already in `processed/`), `mv` it to `processed/`. Update `usage()` text and add a `_log` line on auto-ack. Drift group `radio` enforces byte-identity across all 7. |
| 7× `<loadout>/bin/radio` (region:`body`, `cmd_ack`) | MODIFY | Make idempotent: if `id.md` is in `processed/` already (not in `inbox/`), print a friendly "already processed" line on stdout and exit 0. Today it errors. Drift group `radio`. |
| `tests/radio_read_auto_ack.bats` | CREATE | Pin contract: (a) `radio read <id>` moves file to `processed/`; (b) `radio read --peek <id>` does not; (c) `radio read --no-ack <id>` does not; (d) `radio ack` on an already-processed id exits 0; (e) `radio read` of a message that's already in `processed/` still prints (no move, no error). |
| `tests/radio.bats` (existing) | MODIFY | Audit any test that calls `radio read <id>` and depended on the file *staying* in `inbox/` — update or split into a `--peek` variant. |
| `CHANGELOG.md` | MODIFY | New entry under `[Unreleased]` describing the behavior change. Call it a deliberate UX change, not a bug fix — old behavior was the intended (broken) design. |
| `claude-{gh,jira,notion,local}/commands/{worker,pm,planner}.md` and `kiro-*/agents/*.json` if they document the read/ack pattern | OPTIONAL MODIFY | Trim any guidance that says "after reading, run `radio ack`" — that beat is now implicit. Leave unrelated handoff guidance intact. |

Drift impact: all radio changes are inside the `# region:body` block, already drift-checked across the 7 copies (group `radio` in `tools/check-drift.sh` manifest).

## Verification

1. **CI green:** `./run_tests.sh` and `tools/check-drift.sh` both pass.
2. **Auto-ack happy path:** send a message to a registered role via `radio send`; run `radio read <id>` as that role; confirm (a) the body printed, (b) `~/.task-force/radio/mailbox/<role>/inbox/` no longer contains `<id>.md`, (c) `~/.task-force/radio/mailbox/<role>/processed/` does, (d) subsequent `radio check` returns `(no unread messages)`.
3. **Peek path:** repeat (2) but with `radio read --peek <id>`. File stays in `inbox/`. `radio check` still lists it.
4. **No-ack alias:** repeat (3) with `--no-ack` instead. Same outcome.
5. **Legacy ack idempotent:** read with auto-ack (so file is in `processed/`); then `radio ack <id>`. Exit code 0, prints "already processed" or similar.
6. **Re-read from processed:** after auto-ack, run `radio read --peek <id>` (or `radio read <id>`) again. Body still prints (current `cmd_read` already falls back to `processed/` for the lookup path); no move, no error.
7. **Wrong-role isolation:** send a message to role A. Reading it as role B (`TASK_FORCE_ROLE=B radio read <id>`) errors / no-ops — does not affect role A's inbox. Pin this if not already pinned.
8. **End-to-end (regression of the original bug):** simulate the failure pattern from `feedback-radio-ack-after-handling`. Send a `changes-requested` to a worker role. Worker runs `radio read <id>` (default). Subsequent `radio check` returns `(no unread messages)`. Send a *new* `changes-requested`. Worker's next `check` shows exactly one message — the new one — distinguishable from the now-processed prior.

## Risk notes

- **Breaking change in semantics:** anyone with a workflow that depended on `radio read <id>` leaving the file in `inbox/` (e.g. "let me peek and decide later") will see different behavior. Mitigated by `--peek` / `--no-ack`. Surface this as a `### Changed` entry in CHANGELOG, not buried in `### Fixed`.
- **Test churn:** `tests/radio.bats` and the read/ack pieces of `tests/radio_*.bats` need an audit — the read/ack contract is mutating. Expected churn is small (handful of tests).
- **No drift across loadouts:** all 7 radio copies must move together. Existing `tools/check-drift.sh` `radio` group catches this — CI is the safety net.
- **Comments preserved:** the existing comment block above the wake-up `write-chars` in `cmd_send` (about why we only ever write `"radio check"$'\n'` and not message bodies) is unrelated and must stay intact during the edit. Same goes for the `_silent_exit_if_no_role` hook-gate comments.
- **`--peek` flag name:** `--peek` is the docs-preferred name (terse, evocative). `--no-ack` accepted as a strict-equivalent alias for users who prefer the explicit form. Don't add a third name; aliases compound.

martin-conur:
## Summary

Closes #131.

Collapses `radio read` + `radio ack` into one beat by default. The previous split was a workflow footgun: workers would read a message, act on it, forget to `ack`, and every subsequent `radio check` would re-list the stale entry identically — making PM's silence indistinguishable from PM never pinging.

- `radio read <id>` — prints body AND moves `inbox/<id>.md` → `processed/`
- `radio read --peek <id>` (alias `--no-ack`) — prints body, leaves file in inbox
- `radio ack <id>` — now idempotent: exits 0 with "already processed" if a prior `read` already moved it

Deliberate UX change, not a bug fix — old behavior was the intended (broken) design. Captured under `### Changed` in CHANGELOG, not `### Fixed`.

All 7 loadout radio copies kept in sync (drift group `radio`/`body`).

## Test plan

- [x] `./run_tests.sh` — all 659 tests green (includes new `tests/radio_read_auto_ack.bats`, 8 cases)
- [x] `tools/check-drift.sh` — all 17 groups green
- [x] Auto-ack happy path pinned in `radio_read_auto_ack.bats:(a)`
- [x] `--peek` and `--no-ack` aliases pinned in cases (b) and (c)
- [x] Idempotent ack pinned in case (d)
- [x] Re-read from `processed/` pinned in case (e)
- [x] Wrong-role isolation pinned (sender role A cannot read role B's mailbox)
- [x] Existing `tests/radio.bats` "read prints body; ack moves to processed/" updated to assert the new auto-ack behavior + legacy idempotent ack
- [x] README command surface updated with the new `--peek` row and idempotent ack note

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Changed
+
+- **`radio read <id>` now auto-acks (moves `inbox/` → `processed/`).** The previous `read` + `ack` split was a workflow footgun: workers would `read` a message, act on it, forget to `ack`, and every subsequent `radio check` would re-list the stale entry identically — making PM's silence indistinguishable from PM never having pinged. `read` now collapses both beats by default. An explicit `radio read --peek <id>` (alias: `--no-ack`) is the escape hatch for "I just want to see what's in there without consuming it." `radio ack <id>` is now idempotent: if the message was already moved by a prior `read`, it exits 0 with a friendly "already processed" line so existing `read && ack` scripts and muscle memory keep working. This is a deliberate UX change, not a bug fix — the old behavior was the intended (broken) design. (#131)
+
 ### Added
 
 - **Opt-in auto-submit on radio wake-up for `--auto` workers.** `radio send`'s zellij `write-chars` wake-up has always written `radio check\n` (LF) into the recipient's pane, leaving it sitting in the input box until the user pressed Enter. For workers launched via `task-work --auto` — already an explicit autonomy opt-in, and idle most of their life waiting on PM — that human gate is friction with no upside. `task-work --auto` now exports `TASK_FORCE_AUTO_SUBMIT=1`; `radio register` / `_ensure_session_file` persist that as `AUTO_SUBMIT=1` in the session file; `cmd_send` reads it and switches the wake-up byte to CR (`\r`), which Claude Code's raw-mode input binds to Enter. PM and default (non-`--auto`) workers leave the flag unset and keep the LF wake-up, so a `radio check` ping can never corrupt a partially-typed prompt. (#128)
--- a/README.md
+++ b/README.md
@@ -382,8 +382,9 @@ Role names are addressable strings, not free-form: the PM is `pm`, and each work
 |---------|--------------|
 | `radio send --to <role> --intent <kind> [--pr N] [--issue N] [--body TEXT]` | Send a message (e.g. `--to pm --intent review-requested --pr 42`); body can come from stdin |
 | `radio check`                       | List unread messages addressed to this role |
-| `radio read <id>`                   | Print one message |
-| `radio ack <id>`                    | Mark it acknowledged so it stops showing up in `check` |
+| `radio read <id>`                   | Print one message AND mark it acknowledged (moves `inbox/` → `processed/`) |
+| `radio read --peek <id>`            | Print without acknowledging — for inspection / debugging |
+| `radio ack <id>`                    | Mark it acknowledged (idempotent — no-op if already processed by a prior `read`) |
 | `radio register` / `radio unregister` | Add/remove this tab's session file (`~/.task-force/radio/sessions/<role>.info`) |
 | `radio ready` / `radio busy`        | Toggle this session's `STATE` field — drives the wake-up vs. queue decision on the sender side |
 | `radio orphans`                     | List session files whose heartbeat is >1h stale |
--- a/claude-gh/bin/radio
+++ b/claude-gh/bin/radio
@@ -14,8 +14,8 @@ usage() {
 Usage:
   radio send --to ROLE --intent KIND [--pr N] [--issue N] [--repo PATH] [--body TEXT | < body]
   radio check                            list unread messages for this role
-  radio read   <id>                      print message body
-  radio ack    <id>                      move message to processed/
+  radio read   [--peek|--no-ack] <id>    print body; by default also move inbox/<id>.md -> processed/
+  radio ack    <id>                      move message to processed/ (idempotent: no-op if already processed)
   radio register --role ROLE --tab TAB [--repo PATH] --agent claude|kiro [--loadout NAME]
   radio ready                            mark this role idle (and process pending)
   radio busy                             mark this role busy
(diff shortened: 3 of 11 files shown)


martin-conur:
## Review notes (non-blocking discussion items)

Did a deep pass — implementation is solid (drift clean, CI green, all 8 spec contract points pinned in tests, comments correctly explain the *why*). Three small observations before merge — none are blockers, but want your read on them:

### 1. `ack: ... already processed` → stdout, not stderr

In `cmd_ack`, the friendly idempotent-path message goes to stdout:

```bash
if [[ -f "$processed/${id}.md" ]]; then
  echo "ack: message $id already processed"
  return 0
fi
```

The matching error path 3 lines below uses `>&2`. The asymmetry is intentional (stdout = "exit 0 informational", stderr = "exit 1 error"), but a caller doing `radio ack $id | grep something` would see this line. The project doesn't pipe radio output programmatically anywhere I can find, so I think this is fine — but flagging in case you considered and rejected stderr for a reason.

### 2. Multiple ids permissive

`radio read <id1> <id2>` silently keeps the last id (the while loop just reassigns). Matches existing `cmd_send` permissiveness, so not a regression, but the worker could now lose a message id without warning if a caller has a typo with stray tokens. Strict-mode (error on extra positional args) would be ~3 lines. Worth tightening, or leave consistent with the rest of the file?

### 3. Stale user memory

The spec called out that memory `feedback-radio-ack-after-handling` (in the user's personal memory) effectively becomes obsolete with auto-ack — but that's outside the PR scope (worker can't touch user memory). Just noting so the user updates it post-merge.

---

Once you respond / push any tweaks, I'll either approve as-is or wait for the changes.

martin-conur:
Responses on (1) and (2):

### 1. stdout for `already processed`
Intentional — same reasoning you suspected. The asymmetry is the signal:
- stdout + exit 0 = "informational, ignore me" (idempotent success)
- stderr + exit non-zero = "error, look at me" (unknown id)

Stderr would arguably be a downgrade: it conventionally implies "something went wrong, look at me," and the idempotent path is success. Callers that want quiet have the idiomatic `radio ack $id >/dev/null` available today. No programmatic consumers in-tree to break.

Keeping as stdout.

### 2. Multiple ids permissive
Inherited, not introduced. The pre-PR `cmd_read` was `id="${1:-}"` — same silent drop of extra positional args. The new loop preserves that behavior to stay byte-consistent with the rest of the file (`cmd_send` does the same).

Tightening just `cmd_read` would be a half-measure; "strict everywhere" vs "permissive everywhere" is a clearer principle than touching only the function this PR happened to land in. Happy to file a follow-up issue for a holistic strict-mode pass on radio argv parsing if you want to track it — but I'd keep this PR scoped to the auto-ack collapse.

### 3. Stale memory
Acked — out of worker scope. Will flag at handoff.

martin-conur:
Verified locally on `main` @ 108dabe: (a) `radio read <id>` prints body and moves inbox→processed; (b) `radio read --peek <id>` and `--no-ack` alias both leave the file in inbox; (c) re-acking an already-moved id is idempotent (exit 0, `already processed`), while a genuinely-unknown id still exits 1.

PR #133, opened 2026-05-26
task-work --auto: spawn worker tab without stealing focus (PM stays put)

### LINKED ISSUE #130: task-work --auto: spawn worker tab without stealing focus (PM stays put)
## Problem

Today `zellij action new-tab` always focuses the newly-created tab. When PM dispatches a worker via `task-work --auto`, focus snaps away from PM into the new worker's tab, and the PM has to manually switch back (`zellij action go-to-tab 1`, or Alt+number) before dispatching the next worker. For a PM running 3–4 parallel `--auto` workers from a single seat, this is constant friction.

Verified (zellij 0.44.3, 2026-05-25): `new-tab` has no `--no-focus` / `--background` flag — focus shift is unconditional. But a two-step capture-then-restore *does* work cleanly:

```bash
POS=$(zellij action list-tabs --json | jq -r '.[] | select(.active) | .position')
zellij action new-tab --name "$slug" --cwd "$cwd" -- bash -ic "$cmd"
zellij action go-to-tab $((POS + 1))   # go-to-tab is 1-indexed; .position is 0-indexed
```

Confirmed live: caller tab stays `active=True` after the round-trip. The only artifact is a brief focus flicker (~hundred-ms range) while the new tab is created and we shift back.

For `--auto` workers, the flicker is acceptable: `--auto` is already an explicit "I'm not babysitting this tab" opt-in (see existing `feedback_task_work_auto_default` / `project_radio_idle_is_feature` semantics — combined with PR #129's CR auto-submit, PM never needs to enter the worker's tab in the happy path).

Default `task-work` (no flag) and `task-work --plan` should keep current behavior — the user explicitly wants to land in those tabs.

## Solution

Thread a `stay_on_caller_tab` opt-in from `task-work --auto` into the shared `aw_launch_tab`.

1. **`lib/zellij-tab.sh`**: `aw_launch_tab` gains an optional 4th argument `stay_on_caller_tab` (the literal string `"1"` enables, anything else disables). When enabled AND we're inside zellij AND `jq` is available:
   - Capture the caller's tab position via `zellij action list-tabs --json | jq -r '.[] | select(.active) | .position'` **before** `new-tab`.
   - Run `new-tab` as today (focus shifts to the new tab).
   - Snap focus back: `zellij action go-to-tab $((pos + 1))`.
   - If any of the gates fail (no `$ZELLIJ`, no `jq`, empty position, `go-to-tab` exits non-zero), the function falls through to the legacy focus-shift behavior — never aborts the worker spawn. Match the existing defensive style in this file (compare `aw_zellij_tab_id_by_name`'s empty-stdout-is-fine pattern).

2. **7× `<loadout>/bin/task-work`** (regions `worktree-creation-post-info` — currently two drift groups: `task-work-claude` for the 4 claude-* loadouts and `task-work-kiro` for the 3 kiro-* loadouts): pass `"$AUTO_MODE"` as the 4th arg to both `aw_launch_tab` call sites (the `--no-launch` branch and the regular branch). `AUTO_MODE` is empty when `--auto` was not set, so non-`--auto` paths get the legacy behavior unchanged.

3. **`--plan` interaction**: `AUTO_MODE` and `PLAN_MODE` are mutually exclusive (existing check in `task-work` arg parsing). When `--plan` is set, `AUTO_MODE` is empty → no snap-back → planner tab gets focus. Correct: planner is interactive, you want to land there.

## Files to Create/Modify

| File | Action | Purpose |
|------|--------|---------|
| `lib/zellij-tab.sh` (`aw_launch_tab`) | MODIFY | Accept optional 4th arg `stay_on_caller_tab`. When `"1"`, capture caller tab position before `new-tab`, run `zellij action go-to-tab <pos+1>` after. Defensive: skip cleanly when not in zellij / `jq` missing / position lookup fails — never abort the spawn. |
| 4× `claude-{gh,jira,notion,local}/bin/task-work` (region `worktree-creation-post-info`) | MODIFY | Pass `"$AUTO_MODE"` as 4th arg to both `aw_launch_tab` calls. Drift group `task-work-claude` — all 4 byte-identical. |
| 3× `kiro-{gh,notion,local}/bin/task-work` (region `worktree-creation-post-info`) | MODIFY | Same as above. Drift group `task-work-kiro`. |
| `tests/` (likely a new `tests/zellij_tab_stay_focused.bats` plus extending `tests/helpers/stubs/zellij` to record `go-to-tab` calls if it doesn't already) | CREATE/MODIFY | Pin contract: `stay_on_caller_tab="1"` → recorded sequence is `list-tabs` → `new-tab` → `go-to-tab <pos+1>`. `stay_on_caller_tab=""` or absent → no `go-to-tab` after `new-tab` (legacy). |
| `CHANGELOG.md` | MODIFY | New `### Added` entry under `[Unreleased]`. |

Drift impact: all task-work edits land inside existing drift-checked regions (`tools/check-drift.sh` `task-work-claude` / `task-work-kiro` groups). `lib/zellij-tab.sh` is the single shared file — no drift concern. CI's drift job catches loadout skips.

## Verification

1. **CI green:** `./run_tests.sh` and `tools/check-drift.sh` both pass.
2. **Manual (`--auto` keeps PM focus):** From PM tab, run `task-work test-bg <gh-url> --auto`. New tab appears in tab bar but PM tab stays focused (active=True). Repeat 3 more times in a row from PM — all 4 worker tabs spawn, PM never loses focus.
3. **Manual (default focuses):** `task-work test-default <gh-url>` (no flag) — focus shifts to the new worker tab (current behavior preserved).
4. **Manual (`--plan` focuses):** `task-work test-plan <gh-url> --plan` — focus shifts to the planner tab (planner is interactive; you want to land there).
5. **Manual (non-zellij):** Run `task-work test-noz <gh-url> --auto` from a plain shell outside zellij. Tab spawn still works (or fails the same way it does today); no errors from the new code path.
6. **Combined with #128:** `task-work bg-and-submit <gh-url> --auto`. From PM, `radio send --to worker-task-force-bg-and-submit --intent approved-and-merged --pr 1 --body merged`. Worker tab receives CR wake-up, auto-processes, cleans up — PM never had to leave its tab.

## Risk notes

- **Focus flicker (~100-300ms):** between `new-tab` and `go-to-tab`. If the user is mid-typing in PM during that window, the next keystroke could land in the new tab. Mitigated by: (a) `--auto` is an explicit autonomy opt-in, (b) PM is typically waiting on radio pings, not typing, when dispatching, (c) the window is short. Documented in the diff comment.
- **Backwards compat:** non-`--auto` `task-work` (default + `--plan`) gets the empty `stay_on_caller_tab` arg → falls through to current new-tab behavior → identical legacy UX.
- **Failure modes:** all the gates (`$ZELLIJ` unset, `jq` missing, empty position from list-tabs, `go-to-tab` non-zero) fall through to the legacy focus-shift path. The worker spawn never fails because of the snap-back.
- **Index off-by-one:** `list-tabs --json` reports `position` 0-indexed; `go-to-tab` is 1-indexed. The `+1` is the contract — verified live (2026-05-25) and pinned in the bats test.
- **Zellij version:** verified on 0.44.3 (the project's documented minimum, per README). `list-tabs --json`, `new-tab`, `go-to-tab` are all stable on that version.

martin-conur:
## Summary

- `aw_launch_tab` (`lib/zellij-tab.sh`) gains an optional 4th arg `stay_on_caller_tab`. When `"1"`, capture caller tab position via `zellij action list-tabs --json` before `new-tab`, then snap focus back with `go-to-tab <pos+1>` after. All gates (`$ZELLIJ` unset, `jq` missing, empty position, non-zero `go-to-tab`) fall through to the legacy focus-shift path — the worker spawn is never aborted because the snap-back failed.
- All 7 `<loadout>/bin/task-work` thread `${AUTO_MODE:-}` through to `aw_launch_tab`. `--auto` opts in; default and `--plan` keep current focus-shift behavior (planner is interactive — you want to land there).
- New `tests/zellij_tab_stay_focused.bats` (11 cases) pins the contract: unit tests on `aw_launch_tab` directly, plus 3 integration cases via `claude-gh/bin/task-work` covering `--auto`, default, and `--plan`.

Closes #130.

## Test plan

- [x] `./run_tests.sh` — 662 ok / 0 fail (clean env)
- [x] `tools/check-drift.sh` — 17 group(s) checked, no drift
- [ ] Manual: from PM tab, `task-work test-bg <gh-url> --auto` — new tab appears, PM keeps focus. Repeat 3× in a row.
- [ ] Manual: `task-work test-default <gh-url>` — focus shifts to new tab (legacy).
- [ ] Manual: `task-work test-plan <gh-url> --plan` — focus shifts to planner tab.
- [ ] Manual: non-zellij shell, `task-work test-noz <gh-url> --auto` — no errors, spawn behavior unchanged.


martin-conur:
## Review notes (non-blocking discussion items)

Did a deep pass — implementation is solid (all 4 gates honored, snap-back failure never aborts the spawn, 0/1-index translation pinned in tests, drift clean, CI green). Three small observations before merge:

### 1. `${AUTO_MODE:-}` consistency tweak — half-applied

The diff tightens 3 existing call sites per claude-* task-work file from `$AUTO_MODE` → `${AUTO_MODE:-}`:

```diff
-if [[ -n "$AUTO_MODE" && -n "$PLAN_MODE" ]]; then
+if [[ -n "${AUTO_MODE:-}" && -n "$PLAN_MODE" ]]; then
```

```diff
-[[ -n "$AUTO_MODE" ]] && mode_prefix="--permission-mode auto "
+[[ -n "${AUTO_MODE:-}" ]] && mode_prefix="--permission-mode auto "
```

(a) The file isn't `set -u`, so this changes no runtime behavior. (b) `$PLAN_MODE` is **not** given the same treatment on the same line — so the consistency is half-applied. (c) Scope-adjacent to #130 rather than required by it.

Either revert these to keep the PR tight on #130, or finish the pass to `${PLAN_MODE:-}` too — current state reads as "started a hygiene pass and stopped." Your call.

### 2. Two gate paths not pinned by tests

- **`jq` missing**: the `command -v jq >/dev/null 2>&1` gate has no test that PATH-masks `jq` to confirm fallthrough to legacy behavior.
- **`go-to-tab` non-zero exit**: the `2>/dev/null || true` swallow has no test where the stub errors on `go-to-tab` to confirm the spawn still succeeds.

Both are silent-fallthrough paths — failure mode = "legacy behavior" — so missing tests aren't catastrophic, but they're explicitly called-out defensive paths in the spec and your function header comment. Pinning them would tighten the contract. ~10 lines of bats each.

### 3. Flicker documentation

The function header is honest about the ~100-300ms keystroke window — matches the spec's risk note. Good. No action.

---

Want your read on (1) — that's the only one that may meaningfully affect what lands. (2) is a follow-up if you'd rather scope tightly.

PR #142, opened 2026-06-01
task-init: seed read-only tool allowlist (Read, Grep, Glob, find/ls/cat) into .claude/settings.json

### LINKED ISSUE #141: task-init: seed read-only tool allowlist (Read, Grep, Glob, find/ls/cat) into .claude/settings.json
## Problem

When running `task-work --plan` (or invoking `/planner` directly), the planner agent gets permission prompts for every read-only operation it needs to explore the codebase — `Read`, `Grep`, `Glob`, `Bash(find …)`, `Bash(ls …)`, etc. This is friction for a role whose entire job description (`claude-gh/commands/planner.md`) is "read files, search symbols, grep patterns."

Workers and reviewers hit the same friction — exploration is a cross-role concern, not planner-specific.

## Root cause

Each of the four claude variants of `task-init` (`claude-gh`, `claude-jira`, `claude-notion`, `claude-local`) seeds `.claude/settings.json` `permissions.allow` with tracker-specific reads (`Bash(gh issue view *)` / `mcp__atlassian__*` / `mcp__notion__*`) plus `Bash(radio *)`. None of Claude Code's built-in read tools nor common shell read commands are pre-allowed, so every new project either sits through permission prompts during plan/exploration phases or maintains a per-user `.claude/settings.local.json` workaround.

## Solution

In `_radio_install_hooks()` of each of the four `claude-*/bin/task-init` scripts, extend the jq-merged `permissions.allow` array with these seven entries (alongside the existing tracker-specific entries, which stay where they are since they legitimately differ):

- `Read`
- `Grep`
- `Glob`
- `Bash(find *)`
- `Bash(ls *)`
- `Bash(cat *)`
- `Bash(rg *)`

These are unambiguously read-only and safe across all three roles (planner, worker, reviewer). Mutations (Edit, Write, `Bash(gh issue edit …)`, MCP write tools, etc.) stay confirmation-gated as today.

### Drift enforcement

Wrap **only the new shared array literal lines** in matching `# region:read-only-allow` / `# endregion:read-only-allow` sentinels in each of the four files. Add the region to the `GROUPS_DEFAULT` manifest in `tools/check-drift.sh`:

```
read-only-allow|read-only-allow|claude-gh/bin/task-init claude-jira/bin/task-init claude-local/bin/task-init claude-notion/bin/task-init
```

This guarantees the four variants stay byte-identical for the shared set. The existing per-variant tracker entries remain outside the sentinel since they legitimately differ.

**Implementation note**: jq's `#` introduces a line comment, so the sentinel comments inside the jq program should be a no-op for jq. The worker should validate this with one smoke run before propagating the pattern across all four files; if jq objects, fall back to placing the sentinels as bash comments wrapping the array literal lines just outside the jq invocation (the array contents are the part drift-check actually compares).

## Files to Create/Modify

| File | Action | Purpose |
|------|--------|---------|
| `claude-gh/bin/task-init` | MODIFY | Append 7 new entries to `permissions.allow` array (around lines 292–310); wrap them in `# region:read-only-allow` sentinel |
| `claude-jira/bin/task-init` | MODIFY | Same 7 entries appended (around lines 265–277); same sentinel |
| `claude-notion/bin/task-init` | MODIFY | Same 7 entries appended (around lines 220–233); same sentinel |
| `claude-local/bin/task-init` | MODIFY | Same 7 entries appended (around lines 283–287); same sentinel |
| `tools/check-drift.sh` | MODIFY | Add `read-only-allow` group to `GROUPS_DEFAULT` (inside the existing `# region:default-manifest` block, lines 13–32) |
| `tests/claude_gh_task_init.bats` | MODIFY | Extend the existing `seeds gh read-only patterns into permissions.allow` test (~lines 449–465) to also assert each of the seven new entries via `assert_output --partial` |
| `tests/jira_task_init.bats` | MODIFY | Same assertions added to its `seeds mcp__atlassian__* …` test (~lines 247–260) |
| `tests/claude_notion_task_init.bats` | MODIFY | Same assertions added to its `seeds mcp__notion__* …` test (~lines 231–244) |
| `tests/claude_local_task_init.bats` | MODIFY | Same assertions added to its `seeds Bash(radio *) …` test (~lines 241–274) |

### Reused infrastructure (do not reinvent)

- **jq `unique` merge pattern** in each task-init's `_radio_install_hooks()` already handles idempotence; the new entries piggyback on the existing `+ […] | unique` chain.
- **Region sentinel + drift manifest** at `tools/check-drift.sh:35-89` (`extract_region()` / `check_group()`) is the canonical mechanism; only the manifest needs editing.
- **Bats assertion style** at `tests/claude_gh_task_init.bats:449-465` — use `run jq -r '.permissions.allow[]'` then `assert_output --partial "…"` for each new entry.
- **`refute_output --partial` negative assertions** in the existing tests stay intact; none of the new entries overlap with the mutation patterns they refute.

## Verification

1. **bats suite passes**:
   ```bash
   bats tests/claude_gh_task_init.bats tests/jira_task_init.bats \
        tests/claude_notion_task_init.bats tests/claude_local_task_init.bats
   ```
2. **Drift check passes**:
   ```bash
   tools/check-drift.sh
   ```
   Should print `check-drift: N group(s) checked` with the new `read-only-allow` group counted and no `DRIFT`/`FAIL` lines.
3. **Smoke test in a scratch dir** (validates jq accepts the inline sentinel comments):
   ```bash
   tmp=$(mktemp -d) && cd "$tmp" && git init -q
   /full/path/to/claude-gh/bin/task-init
   jq '.permissions.allow' .claude/settings.json
   ```
   Output should include all seven new entries plus the existing `Bash(gh …)` and `Bash(radio *)` entries. Repeat for the other three variants in fresh scratch dirs.
4. **Idempotency**: re-running `task-init <variant> --force` in the same scratch dir leaves `.permissions.allow | length` unchanged (existing idempotency tests in each suite will exercise the new entries automatically).
5. **End-to-end (manual, optional)**: in a fresh worktree, launch `claude --permission-mode plan` and have the planner explore the codebase. Read/Grep/Glob/find/ls operations should not trigger prompts; mutating tools (Edit, Write, `Bash(gh issue edit …)`) still should.

## Scope

- Claude loadouts only — Kiro has its own permission model (`allowedTools` inside agent JSON files), so any equivalent change there belongs in a separate issue if needed.

## Out of scope

- Backfilling already-initialized projects' `.claude/settings.json` (users re-run `task-init <variant> --force` themselves to pick up new seeds).
- Kiro variants (separate change, separate issue if pursued).
- Dogfooding this repo's own `.claude/settings.json` — re-running `task-init claude-gh --force` here will land the seven new entries naturally; no manual edit needed.

martin-conur:
## Summary

- Append seven shared read-only entries to `permissions.allow` in all four claude-* `task-init` variants: `Read`, `Grep`, `Glob`, `Bash(find *)`, `Bash(ls *)`, `Bash(cat *)`, `Bash(rg *)`. Eliminates permission prompts for routine codebase exploration by planner / worker / reviewer roles; mutations stay confirmation-gated.
- Wrap the shared block in `# region:read-only-allow` sentinels and add the group to `tools/check-drift.sh`'s `GROUPS_DEFAULT` manifest so the four variants stay byte-identical.
- Extend the existing `seeds … permissions.allow` test in each of the four bats files with `assert_output --partial` for every new entry.

Closes #141

## Test plan
- [x] `bats tests/claude_gh_task_init.bats tests/jira_task_init.bats tests/claude_notion_task_init.bats tests/claude_local_task_init.bats` — 111/111 ok
- [x] `bats tests/drift_check.bats` — 7/7 ok
- [x] `tools/check-drift.sh` — `check-drift: 19 group(s) checked`
- [x] Smoke test: `task-init` in a scratch dir for each of the four variants produces a `.permissions.allow` containing all 7 shared entries plus its tracker-specific reads
- [ ] Manual: launch `claude --permission-mode plan` in a fresh worktree; confirm Read/Grep/Glob/find/ls do not prompt while Edit/Write/`gh issue edit` still do

### DIFF
--- a/claude-gh/bin/task-init
+++ b/claude-gh/bin/task-init
@@ -305,7 +305,16 @@ _radio_install_hooks() {
           "Bash(gh label list *)",
           "Bash(gh auth status)",
           "Bash(gh repo view *)",
-          "Bash(radio *)"
+          "Bash(radio *)",
+# region:read-only-allow
+          "Read",
+          "Grep",
+          "Glob",
+          "Bash(find *)",
+          "Bash(ls *)",
+          "Bash(cat *)",
+          "Bash(rg *)"
+# endregion:read-only-allow
         ] | unique
       )
     ' "$settings_file" >"$tmp"
--- a/claude-gh/bin/task-init
+++ b/claude-gh/bin/task-init
@@ -103,6 +103,10 @@ LIB_DIR="$SCRIPT_DIR/../../lib"
 . "$LIB_DIR/install-file.sh"
 # shellcheck source=lib/preserve-placeholders.sh
 . "$LIB_DIR/preserve-placeholders.sh"
+# region:require-jq-source
+# shellcheck source=lib/require-jq.sh
+. "$LIB_DIR/require-jq.sh"
+# endregion:require-jq-source
 
 if [[ ! -f "$TEMPLATE" ]]; then
   echo "Error: template not found at $TEMPLATE"
@@ -240,11 +244,9 @@ _radio_install_hooks() {
   local settings_file="$REPO_ROOT/.claude/settings.json"
   local loadout="claude-gh"
 
-  if ! command -v jq >/dev/null 2>&1; then
-    echo "Error: jq is required to merge radio hooks into $settings_file" >&2
-    echo "       Install jq (https://jqlang.github.io/jq/) and re-run task-init." >&2
-    exit 1
-  fi
+# region:require-jq-call
+  require_jq_1_6 || exit 1
+# endregion:require-jq-call
 
   # \${TASK_FORCE_LOADOUT:-$loadout} lets per-role launchers (e.g. task-reviewer
   # exporting TASK_FORCE_LOADOUT=reviewer) override the LOADOUT= field in the
--- a/claude-jira/bin/task-init
+++ b/claude-jira/bin/task-init
@@ -89,6 +89,10 @@ LIB_DIR="$SCRIPT_DIR/../../lib"
 . "$LIB_DIR/install-file.sh"
 # shellcheck source=lib/preserve-placeholders.sh
 . "$LIB_DIR/preserve-placeholders.sh"
+# region:require-jq-source
+# shellcheck source=lib/require-jq.sh
+. "$LIB_DIR/require-jq.sh"
+# endregion:require-jq-source
 
 if [[ ! -f "$TEMPLATE" ]]; then
   echo "Error: template not found at $TEMPLATE"
@@ -209,11 +213,9 @@ _radio_install_hooks() {
   local settings_file="$REPO_ROOT/.claude/settings.json"
   local loadout="claude-jira"
 
-  if ! command -v jq >/dev/null 2>&1; then
-    echo "Error: jq is required to merge radio hooks into $settings_file" >&2
-    echo "       Install jq (https://jqlang.github.io/jq/) and re-run task-init." >&2
-    exit 1
-  fi
+# region:require-jq-call
+  require_jq_1_6 || exit 1
+# endregion:require-jq-call
 
   # \${TASK_FORCE_LOADOUT:-$loadout} lets per-role launchers (e.g. task-reviewer
   # exporting TASK_FORCE_LOADOUT=reviewer) override the LOADOUT= field in the
--- a/claude-jira/bin/task-init
+++ b/claude-jira/bin/task-init
@@ -272,7 +272,16 @@ _radio_install_hooks() {
           "mcp__atlassian__getIssueLinkTypes",
           "mcp__atlassian__lookupJiraAccountId",
           "mcp__atlassian__atlassianUserInfo",
-          "Bash(radio *)"
+          "Bash(radio *)",
+# region:read-only-allow
+          "Read",
+          "Grep",
+          "Glob",
+          "Bash(find *)",
+          "Bash(ls *)",
+          "Bash(cat *)",
+          "Bash(rg *)"
+# endregion:read-only-allow
         ] | unique
       )
     ' "$settings_file" >"$tmp"
--- a/claude-local/bin/task-init
+++ b/claude-local/bin/task-init
@@ -73,6 +73,10 @@ TEMPLATE="$SCRIPT_DIR/../steering/local-workflow.example.md"
 LIB_DIR="$SCRIPT_DIR/../../lib"
 # shellcheck source=lib/install-file.sh
 . "$LIB_DIR/install-file.sh"
+# region:require-jq-source
+# shellcheck source=lib/require-jq.sh
+. "$LIB_DIR/require-jq.sh"
+# endregion:require-jq-source
 
 if [[ ! -f "$TEMPLATE" ]]; then
   echo "Error: template not found at $TEMPLATE"
@@ -234,11 +238,9 @@ _radio_install_hooks() {
   local settings_file="$REPO_ROOT/.claude/settings.json"
   local loadout="claude-local"
 
-  if ! command -v jq >/dev/null 2>&1; then
-    echo "Error: jq is required to merge radio hooks into $settings_file" >&2
-    echo "       Install jq (https://jqlang.github.io/jq/) and re-run task-init." >&2
-    exit 1
-  fi
+# region:require-jq-call
+  require_jq_1_6 || exit 1
+# endregion:require-jq-call
 
   # \${TASK_FORCE_LOADOUT:-$loadout} lets per-role launchers (e.g. task-reviewer
   # exporting TASK_FORCE_LOADOUT=reviewer) override the LOADOUT= field in the
--- a/claude-local/bin/task-init
+++ b/claude-local/bin/task-init
@@ -282,7 +282,16 @@ _radio_install_hooks() {
     | .permissions.allow //= []
     | .permissions.allow = (
         .permissions.allow + [
-          "Bash(radio *)"
+          "Bash(radio *)",
+# region:read-only-allow
+          "Read",
+          "Grep",
+          "Glob",
+          "Bash(find *)",
+          "Bash(ls *)",
+          "Bash(cat *)",
+          "Bash(rg *)"
+# endregion:read-only-allow
         ] | unique
       )
     ' "$settings_file" >"$tmp"
--- a/claude-notion/bin/task-init
+++ b/claude-notion/bin/task-init
@@ -228,7 +228,16 @@ _radio_install_hooks() {
           "mcp__notion__notion-get-users",
           "mcp__notion__notion-get-user",
           "mcp__notion__notion-get-self",
-          "Bash(radio *)"
+          "Bash(radio *)",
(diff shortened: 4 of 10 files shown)


martin-conur:
## Review: task-init read-only allow-list (#141)

### Spec compliance

PR delivers everything the issue asked for:

| Deliverable | Status |
|---|---|
| 7 new entries in `claude-gh/bin/task-init` with `read-only-allow` sentinel | ✓ |
| 7 new entries in `claude-jira/bin/task-init` with sentinel | ✓ |
| 7 new entries in `claude-notion/bin/task-init` with sentinel | ✓ |
| 7 new entries in `claude-local/bin/task-init` with sentinel | ✓ |
| `read-only-allow` group added to `tools/check-drift.sh` `GROUPS_DEFAULT` inside `# region:default-manifest` | ✓ |
| Bats assertions extended in all 4 test files | ✓ |
| Existing `refute_output` mutation guards preserved | ✓ |
| Regions are byte-identical across the four variants | ✓ (drift check confirms) |

The sentinel comment placement inside the jq heredoc is sound: jq treats `#` as a line comment, so the array parses correctly, and `extract_region`'s awk reads raw file bytes so it matches the column-0 sentinel lines exactly regardless.

The unchecked test-plan item ("Manual: launch `claude --permission-mode plan`…") is explicitly marked optional in the spec; no issue there.

### Code-review findings

**Nit — jq 1.6 required for `#` comments; no version guard**

The four task-init scripts now embed `# region:read-only-allow` / `# endregion:read-only-allow` inside the jq program string (e.g. `claude-local/bin/task-init` lines ~286/294). jq `#` comment support was introduced in jq 1.6 (2018); jq 1.5 would error out:

```
jq: error (at <stdin>:N): syntax error, unexpected '#'
```

causing `task-init` to exit before writing `settings.json`. There is no version check in any of the four scripts, and the CI workflow does not pin a minimum jq version (it uses whatever ships with `ubuntu-latest` / `macos-latest`).

In practice, virtually every 2019+ distro ships jq 1.6+, so real-world breakage is unlikely. The author's smoke test confirmed it works in their environment. This is a nit, not a blocker — a one-liner guard or a note in the README/CONTRIBUTING prerequisites would close the gap if cross-platform portability matters.

No other correctness, security, or logic issues found. The `| unique` idempotency, the drift-check integration, and the test assertion coverage are all correct.

### Verdict

**clean-with-nits** — one low-severity compatibility nit (jq 1.6 assumed, not enforced). No blockers. PM can merge at their discretion.

martin-conur:
## Changes requested — address reviewer nit before merge

Reviewer flagged one low-severity finding (jq 1.6+ assumed by the inline `#` sentinels, no version guard). Let's close that gap before merging.

### Ask

Add a jq version guard at the top of `_radio_install_hooks()` in each of the four `claude-*/bin/task-init` scripts — before the heredoc that contains the sentinel comments. Something like:

```bash
_require_jq_1_6() {
  local v
  v=$(jq --version 2>/dev/null | sed 's/^jq-//') || {
    echo "Error: jq is required (>= 1.6)" >&2
    return 1
  }
  # Compare X.Y; jq versions are dot-major.minor[.patch]
  local maj min
  IFS='.' read -r maj min _ <<<"$v"
  if (( maj < 1 || (maj == 1 && min < 6) )); then
    echo "Error: jq >= 1.6 required (have $v) — task-init uses '#' comments inside the jq program (introduced in 1.6)" >&2
    return 1
  fi
}
```

Then call it once before the heredoc invocation that fails on < 1.6.

### Scope

- Add the guard to all four claude variants (the kiro variants don't embed the same sentinels, so they're unaffected — but if the helper makes more sense as a shared `lib/` function rather than four copies, that's fine too; either layout works for me).
- Drift-check this helper if you DRY it across the four variants (a new region, or share via `lib/` — your call).
- No test changes strictly required, but a single bats case asserting the guard exits non-zero when `jq` is stubbed to print `jq-1.5` would tie it down nicely.

### Out of scope

- README prerequisite update is welcome but optional — the runtime guard is the load-bearing fix.
- Don't go beyond the version guard. The rest of the PR is clean.

Ping me on `re-review-requested` when ready.

martin-conur:
## Re-review: task-init read-only allow-list + jq guard (#141)

_This re-review covers the follow-up commit (`89f7695`) that addresses the jq 1.6 version-guard nit from the first review round._

---

### Spec compliance

All original issue deliverables remain present, plus the changes-requested items are fully addressed:

| Deliverable | Status |
|---|---|
| 7 read-only entries + `read-only-allow` sentinel in all 4 variants | ✓ |
| `read-only-allow` group in `tools/check-drift.sh` `GROUPS_DEFAULT` | ✓ |
| Bats `assert_output --partial` assertions in all 4 test files | ✓ |
| Existing `refute_output` mutation guards preserved | ✓ |
| Regions byte-identical across the four variants (drift-checked) | ✓ |
| jq 1.6 version guard added (`lib/require-jq.sh`, `require_jq_1_6`) | ✓ |
| Guard DRY'd into shared `lib/` — drift-checked via `require-jq-source` / `require-jq-call` regions | ✓ |
| Single bats test for jq < 1.6 guard (stubs jq to print `jq-1.5`) | ✓ |

The spec comment said "a single bats case" — having it in `claude_gh_task_init.bats` is correct since the guard lives in shared `lib/require-jq.sh`.

---

### Code-review findings

**Nit 1 — `[[ -z "$min" ]]` rejects a future single-component major version (e.g. `jq-2`)**

`lib/require-jq.sh:25`

```bash
IFS='.' read -r maj min _ <<<"$v"
maj=${maj%%[!0-9]*}
min=${min%%[!0-9]*}
if [[ -z "$maj" || -z "$min" ]] || (( maj < 1 || (maj == 1 && min < 6) )); then
```

If `jq --version` outputs `jq-2` (no dot), `min` ends up empty, the `[[ -z "$min" ]]` guard fires, and jq 2 is rejected as if it were too old — before the arithmetic comparison even runs. jq 2 does not exist today, but the guard is structurally wrong for that input. Fix: treat empty `min` as `0` rather than an error:

```bash
maj=${maj:-0}
min=${min:-0}
# then just the arithmetic check:
if (( maj < 1 || (maj == 1 && min < 6) )); then
```

**Nit 2 — `if ! v=$(jq --version 2>/dev/null)` can misfire on non-standard builds**

`lib/require-jq.sh:17`

A jq shim that prints a version string to stdout but exits non-zero (e.g. some Homebrew wrappers return 2) would set `v` to the version string yet still enter the error branch, printing `"jq --version' failed"` even though jq is present and compatible. Safer pattern:

```bash
v=$(jq --version 2>/dev/null)
if [[ -z "$v" ]]; then
  echo "Error: 'jq --version' produced no output …" >&2
  return 1
fi
```

Neither finding is a blocker — the first is hypothetical until jq 2 ships, and the second only triggers on broken/custom builds.

---

### Verdict

**clean-with-nits** — the previous blocker (jq 1.6 guard) is fully resolved with a clean shared-library approach. Two low-severity nits in the version-parsing logic; neither warrants blocking the merge. PM can merge at their discretion.

PR #150, opened 2026-06-02
Radio session corruption: unregister cascade → re-seed loses TAB_ID + LOADOUT

### LINKED ISSUE #140: Radio session corruption: unregister cascade → re-seed loses TAB_ID + LOADOUT (breaks --auto wake-up)
## Problem

Long-running worker sessions degrade their own session file: `TAB_ID` becomes empty, `LOADOUT` becomes `unknown`. Once that happens, `cmd_send`'s `--auto` wake-up via `zellij write-chars --tab-id $TAB_ID` silently no-ops, and queued radio messages sit unread in the worker's inbox forever — PM's `changes-requested` pings hit a dead zone. Today I had to manually nudge the worker on PR #139 twice because of this.

Reproduced 100% on the worker for `task/reviewer-redesign-138`. Same pattern visible on every long-lived worker session.

### Bug A — Unregister cascade

`radio.log` shows the trigger sequence for the broken worker:

```
13:27:43Z register role=worker-task-force-reviewer-redesign-138 tab=reviewer-redesign-138 tab_id=4 agent=claude loadout=claude-gh   ← initial: correct
13:47:41Z unregister role=worker-task-force-reviewer-redesign-138
13:47:43Z unregister ...                                                                                                            ← 2nd within 2s
13:47:43Z unregister ...                                                                                                            ← 3rd
13:47:44Z ensure_session: re-seeded ... (file was missing) tab_id=                                                                  ← TAB_ID lost
13:47:44Z _rename_tab: no tab_id for role ... — skipping
13:47:45Z unregister ...                                                                                                            ← cascade continues
13:47:45Z unregister ...
13:47:46Z unregister ...
... 20+ unregister calls in ~30 seconds ...
```

`cmd_unregister`'s skip-list (`claude-gh/bin/radio:367` ish) only matches `reason == clear` or `resume` from the SessionEnd hook payload. Whatever Claude Code is firing here doesn't carry those reasons (or carries them malformed) and slips through — repeatedly. The protection added for #109 doesn't catch this pattern.

The cascade itself is a separate bug — *something* is invoking SessionEnd 20+ times for a single still-alive worker.

### Bug B — `_ensure_session_file` re-seed loses `TAB_ID`

When `busy`/`ready` runs after `unregister` deleted the file, `_ensure_session_file` (line 153–186) rebuilds it. The rebuild path does:

```bash
tab_id=$(_zellij_tab_id_by_name "$ZELLIJ_TAB" || true)
```

By this point `_rename_tab` has painted the visible tab name to `▶️ reviewer-redesign-138` or `⏸️ reviewer-redesign-138` (the busy/idle emoji prefix). `$ZELLIJ_TAB` is the bare slug `reviewer-redesign-138`. `_zellij_tab_id_by_name` does an exact-match against the visible name → no match → returns empty → `TAB_ID=` written empty.

`cmd_register` (line 318–325) has explicit `#117` protection for this exact case: it preserves an existing non-empty `TAB_ID` when its own lookup fails. But that protection only runs in `cmd_register`, not in `_ensure_session_file` — even though both face the identical emoji-prefix problem.

### Bug C — `_ensure_session_file` re-seed loses `LOADOUT`

Same function, line 170–172:

```bash
printf '...LOADOUT=%s\n' ... "${TASK_FORCE_LOADOUT:-unknown}"
```

`_ensure_session_file` runs inside the `radio busy` / `radio ready` hook subshell — which inherits Claude Code's hook env, **not** `task-work`'s launch-time exports. `$TASK_FORCE_LOADOUT` is unset there, so it falls through to the literal string `unknown`.

`cmd_register`'s payload uses `"$loadout"` (the CLI arg from `--loadout claude-gh`), which is correct. `_ensure_session_file`'s payload uses the env var, which is wrong — there's no CLI arg to fall back to, but the existing file *did* have the right value before it was deleted.

## Symptom matrix

| File state | After `cmd_register` | After cascade-`unregister` + `_ensure_session_file` |
|---|---|---|
| `TAB_ID` | `4` (correct) | `` (lost) |
| `LOADOUT` | `claude-gh` (correct) | `unknown` (lost) |
| Wake-up via `cmd_send`? | works (CR-submit to `--tab-id 4`) | silently no-ops (no `--tab-id`) |

Result: `--auto` workers stop being auto. PM thinks ping was delivered; worker never sees it until a human keypress fires Stop hook → `radio check` surfaces the queued message.

## Fix paths

Open question is whether to fix (A) — eliminate the cascade — or just harden (B) and (C) so the cascade can't corrupt data. Probably both, but A is the underlying root cause.

### Fix A: stop the unregister cascade

The 20+ unregister calls in 30s aren't normal. Possibilities to investigate:
- SessionEnd hook firing on intra-session events whose `reason` isn't in the skip-list (`compact`? a new reason Claude Code added? no reason at all?)
- Some retry loop in Claude Code's hook engine
- Our own `SessionEnd` hook listed multiple times in `.claude/settings.json`

Concrete action: instrument `cmd_unregister` to log the *full hook payload* (not just `reason`) for a few worker runs, see what's actually being sent. Then expand the skip-list or fix the upstream trigger.

### Fix B: `_ensure_session_file` should reuse `cmd_register`'s `#117` protection

If the zellij lookup misses, fall back to preserving the existing `TAB_ID` *from a sidecar* — since by definition the session file itself doesn't exist yet (that's why we're re-seeding), the sidecar would need to be somewhere else. Options:

- Strip emoji prefix from the visible tab name before matching (cheap, fixes the lookup itself):
  ```bash
  # In _zellij_tab_id_by_name, match against both literal $name AND
  # the regex /(▶️|⏸️|⏳)? ?$name/
  ```
- Persist `TAB_ID` to a per-role sidecar file at `register` time (`<sessions>/<role>.tab_id`), survives session-file deletion. `_ensure_session_file` reads from sidecar if zellij lookup misses.

The emoji-prefix-stripping approach is simpler and fixes the root cause symmetrically across `cmd_register` *and* `_ensure_session_file` *and* `_rename_tab`'s tab-name lookups. Recommend that.

### Fix C: `_ensure_session_file` should not depend on `$TASK_FORCE_LOADOUT` env

Two options:
- Same sidecar-file approach as Fix B (persist `LOADOUT` to `<sessions>/<role>.loadout` at register time).
- OR: derive `LOADOUT` from a stable source like the project's `.claude/settings.json` (parse the `--loadout` arg out of the SessionStart hook command). Slightly icky but no extra files.

Sidecar is cleaner — same mechanism as Fix B.

## Why P1

- Breaks the `--auto` worker promise (hands-off dispatch). I had to manually keypress nudge twice today on a single worker.
- Affects every long-running worker — anyone using `task-work --auto` on a non-trivial task will hit this within an hour or two.
- Compounds with the `task-reviewer` redesign (#138 / PR #139): the new dispatch-style reviewer also relies on radio wake-up for verdict delivery. If reviewers age past a SessionEnd cycle before posting their verdict, PM never hears back.

## Reproduction

```bash
# In any task-force repo:
task-work test-radio-corruption "https://github.com/owner/repo/issues/1" --auto
# Wait for SessionEnd to fire at least once (any /compact / /clear / Stop event).
cat ~/.task-force/radio/sessions/worker-task-force-test-radio-corruption.info
# → TAB_ID= empty, LOADOUT=unknown
tail ~/.task-force/radio/log | grep test-radio-corruption
# → see register-with-good-values → unregister cascade → re-seed with lost values
```

## Out of scope

- The broader Claude Code SessionEnd semantics (whether `reason=compact` should be added to the skip-list is one part of Fix A, but the *cascade* itself is the bigger concern).
- Auto-recovery (a watchdog that detects degraded sessions and re-registers) — symptom treatment, not cure.
- Switching from session-file-as-state to a different store (DB, lockfile, etc.) — too big for this issue.

## References

- Issue #109 (Done): "Radio session lifecycle bug: unregister loop wipes session file while worker is still alive" — earlier attempt at this same bug. Cascade evidently still possible.
- Issue #117 (Done): "task-done tab leak: TAB_ID gets clobbered when radio re-registers on session re-init (--auto)" — added the `cmd_register` `#117` protection that `_ensure_session_file` is missing.
- Issue #94 (Done): "Auto-unregister radio sessions on session/worker end" — added the `SessionEnd → unregister` hook that's now cascading.
- Issue #131 (Done): `radio read` auto-acks — related lifecycle hardening.
- `claude-gh/bin/radio:153-186` — `_ensure_session_file` (the re-seed path).
- `claude-gh/bin/radio:280-355` — `cmd_register` (the initial register path with `#117` protection).
- `claude-gh/bin/radio:188-208` — `_update_state` (busy/ready — *this part actually works correctly*, preserves fields via awk).
- `~/.task-force/radio/log` — the smoking gun showing the cascade.

martin-conur:
## Summary

- **Bug B (TAB_ID lost on re-seed):** `_zellij_tab_id_by_name` now matches both the bare slug and the emoji-prefixed visible names (⏸️, ▶️, ❓︎) that `_rename_tab` paints. Symmetric fix across `cmd_register`, `_ensure_session_file`, and `_rename_tab`'s lookups.
- **Bug C (LOADOUT lost on re-seed):** `cmd_register` writes the loadout name to a sidecar (`<sessions>/<role>.loadout`); `_ensure_session_file` reads it during a mid-life re-seed. `cmd_unregister` sweeps the sidecar on real exit; the `clear|resume` skip-list leaves it intact, matching `.info` semantics.
- **Bug A (the cascade itself):** instrumentation only — `cmd_unregister` now logs the full hook payload (one-line) when it proceeds past the skip-list, so the next time the cascade fires we have data on what reason / non-reason is driving the 20+-in-30s `SessionEnd` storm.

All three were chained: a cascade wiped `<role>.info` → `_ensure_session_file` re-seeded → `TAB_ID=` empty + `LOADOUT=unknown` → `cmd_send` couldn't wake the worker → `--auto` workers stopped being auto.

Synced across all seven loadouts (claude-gh / claude-jira / claude-notion / claude-local / kiro-gh / kiro-local / kiro-notion).

Closes #140

## Test plan

- [x] `./run_tests.sh radio_lifecycle` — 25/25 pass, including 12 new tests for the #140 scenarios
- [x] `./run_tests.sh` — 770/770 pass (no regressions in radio, task-work, task-pm, task-reviewer, task-done, drift check, etc.)
- [x] Drift check (`tools/check-drift.sh` runs as part of the suite) confirms all 7 `radio` scripts are byte-identical in the `body` region
- [ ] Manual smoke: spawn `task-work test-radio-140 ... --auto`, trigger a few `/compact` / `/clear` cycles, confirm `~/.task-force/radio/sessions/<role>.info` keeps a non-empty `TAB_ID=` and `LOADOUT=claude-gh` across the lifecycle
- [ ] Manual smoke: tail `~/.task-force/radio/log` during the above and confirm any unregister calls that proceed past the skip-list log their full payload (Bug A instrumentation lets us follow up on the cascade root cause)

### DIFF
--- a/claude-gh/bin/radio
+++ b/claude-gh/bin/radio
@@ -200,9 +200,20 @@ _ensure_session_file() {
   local loadout=""
   local sidecar
   sidecar=$(_loadout_sidecar "$TASK_FORCE_ROLE")
-  if [[ -f "$sidecar" ]]; then
-    loadout=$(<"$sidecar")
-  fi
+  # TOCTOU-safe read (#150 review): the sidecar can be unlinked by a
+  # concurrent cmd_unregister between any pre-check and the read — the
+  # unregister cascade documented in #140 is exactly that race window.
+  # `cat … 2>/dev/null || true` collapses three failure modes into "empty
+  # string" without tripping set -e: file absent, file unlinked mid-read,
+  # file unreadable. The env-var fallback below handles the empty case.
+  #
+  # Why not the obvious `loadout=$(<"$sidecar" || true)` the reviewer
+  # suggested: bash's `<file` is a command-substitution-only shorthand, not
+  # a real command, and `<file || true` short-circuits such that the read
+  # never returns the file's contents — so the sidecar value would be lost
+  # even on the happy path. Verified with `bash -c 'echo x>/tmp/f; v=$(</tmp/f || true); echo "[$v]"'`
+  # printing `[]`. cat is the unambiguous form.
+  loadout=$(cat "$sidecar" 2>/dev/null || true)
   [[ -n "$loadout" ]] || loadout="${TASK_FORCE_LOADOUT:-unknown}"
 
   local repo
--- a/claude-gh/bin/radio
+++ b/claude-gh/bin/radio
@@ -111,13 +111,29 @@ _strip_tab_prefix() {
 # Resolve a zellij tab's stable id from its visible name. Empty stdout means
 # "lookup failed" (zellij not reachable, jq absent, name not found) — callers
 # must treat that as a hard skip so we never operate on the wrong tab (#102).
+#
+# Emoji-prefix awareness (#140): _rename_tab paints the visible tab name to
+# "⏸️ <slug>" / "▶️ <slug>" / "❓︎ <slug>" reflecting STATE. Callers always
+# query by the bare slug ($ZELLIJ_TAB / cmd_register's --tab arg) — that's
+# the stable identifier captured at pane creation and never mutates. So a
+# literal name match against the painted name misses on every flip past
+# idle. We compare against both the bare name AND any of the three emoji-
+# prefixed variants so the lookup survives _rename_tab's repaints. This is
+# the symmetric fix for #117 (cmd_register) AND the same loss in
+# _ensure_session_file's re-seed path (Bug B in #140).
 _zellij_tab_id_by_name() {
   local target="$1"
   [[ -n "$target" ]] || return 0
   command -v jq >/dev/null 2>&1 || return 0
   zellij action list-tabs --json 2>/dev/null \
-    | jq -r --arg n "$target" \
-        '[.[] | select(.name == $n) | .tab_id] | .[0] // empty' 2>/dev/null
+    | jq -r --arg n "$target" '
+        [.[] | select(
+          .name == $n
+          or .name == ("⏸️ " + $n)
+          or .name == ("▶️ " + $n)
+          or .name == ("❓︎ " + $n)
+        ) | .tab_id] | .[0] // empty
+      ' 2>/dev/null
 }
 
 # Pick a writable pane on the given tab. Prefer the tab's own focused pane
@@ -137,6 +153,18 @@ _zellij_pane_in_tab() {
 
 _session_file() {
   printf '%s/%s.info' "$SESSIONS_DIR" "$1"
+}
+
+# Sidecar for the per-role loadout name (#140). cmd_register writes it from
+# its --loadout CLI arg; _ensure_session_file reads from it during a re-seed.
+# Separate from the main session file so it survives cmd_unregister's `rm -f`
+# of <role>.info — by definition, we re-seed because the .info file was just
+# wiped, so we can't read the previous LOADOUT= line out of it. The bare env
+# var ($TASK_FORCE_LOADOUT) isn't available either: hook subshells inherit
+# Claude Code's env, not task-work's launch-time exports — that fallback to
+# the literal string "unknown" was Bug C in #140.
+_loadout_sidecar() {
+  printf '%s/%s.loadout' "$SESSIONS_DIR" "$1"
 }
 
 _inbox_dir()    { printf '%s/%s/inbox'     "$MAILBOX_DIR" "$1"; }
@@ -163,13 +191,27 @@ _ensure_session_file() {
     tab_id=$(_zellij_tab_id_by_name "$ZELLIJ_TAB" || true)
   fi
 
+  # LOADOUT recovery (#140 Bug C): the hook subshell doesn't inherit
+  # $TASK_FORCE_LOADOUT, so the previous fallback to "${TASK_FORCE_LOADOUT:-unknown}"
+  # wrote the literal string "unknown" on every re-seed. Read from the
+  # sidecar that cmd_register persisted at register time. Fall back to the
+  # env var only as a last resort (covers test harnesses that exercise
+  # _ensure_session_file without going through register first).
+  local loadout=""
+  local sidecar
+  sidecar=$(_loadout_sidecar "$TASK_FORCE_ROLE")
+  if [[ -f "$sidecar" ]]; then
+    loadout=$(<"$sidecar")
+  fi
+  [[ -n "$loadout" ]] || loadout="${TASK_FORCE_LOADOUT:-unknown}"
+
   local repo
   repo=$(git rev-parse --show-toplevel 2>/dev/null || echo "")
 
   {
     printf 'ROLE=%s\nTAB=%s\nTAB_ID=%s\nREPO=%s\nSTATE=idle\nLAST_HEARTBEAT=%s\nAGENT=%s\nLOADOUT=%s\n' \
       "$TASK_FORCE_ROLE" "$ZELLIJ_TAB" "$tab_id" "$repo" "$(_now)" \
-      "${TASK_FORCE_AGENT:-claude}" "${TASK_FORCE_LOADOUT:-unknown}"
+      "${TASK_FORCE_AGENT:-claude}" "$loadout"
     # Preserve the --auto opt-in across self-heal so cmd_send keeps using CR
     # for wake-up after an intra-session re-seed (#128). Use `if` rather than
     # `&&` so the brace group exits 0 when the flag is absent — otherwise
@@ -181,7 +223,7 @@ _ensure_session_file() {
   } | _atomic_write "$f"
 
   mkdir -p "$(_inbox_dir "$TASK_FORCE_ROLE")" "$(_processed_dir "$TASK_FORCE_ROLE")"
-  _log "ensure_session: re-seeded $TASK_FORCE_ROLE (file was missing) tab_id=$tab_id"
+  _log "ensure_session: re-seeded $TASK_FORCE_ROLE (file was missing) tab_id=$tab_id loadout=$loadout"
 }
 
 # Update STATE=… and LAST_HEARTBEAT=… on the current session file.
@@ -348,6 +390,16 @@ cmd_register() {
   } | _atomic_write "$f"
 
   mkdir -p "$(_inbox_dir "$role")" "$(_processed_dir "$role")"
+
+  # Persist the loadout name to a sidecar so _ensure_session_file can recover
+  # it during a mid-life re-seed (#140 Bug C). Skipping on empty values keeps
+  # the sidecar's mere existence equivalent to "we have a real loadout to
+  # restore" — otherwise a re-seed could read an empty sidecar and clobber
+  # the env-var fallback that test paths still rely on.
+  if [[ -n "$loadout" ]]; then
+    printf '%s' "$loadout" | _atomic_write "$(_loadout_sidecar "$role")"
+  fi
+
   _log "register role=$role tab=$tab tab_id=$tab_id agent=$agent loadout=$loadout"
 
   # First paint of the idle prefix without waiting for a Stop hook cycle (#95).
@@ -374,7 +426,7 @@ cmd_unregister() {
   # in that case. We only filter when both (a) stdin is not a terminal and
   # (b) a parseable `reason` is present.
   if [[ ! -t 0 ]]; then
-    local payload reason
+    local payload reason payload_oneline
     payload=$(cat 2>/dev/null || true)
     if [[ -n "$payload" ]] && command -v jq >/dev/null 2>&1; then
       reason=$(printf '%s' "$payload" | jq -r '.reason // empty' 2>/dev/null || true)
@@ -385,11 +437,25 @@ cmd_unregister() {
           ;;
       esac
     fi
+    # Instrumentation for #140: when the hook fires with a payload that
+    # *doesn't* match the skip-list, capture the full payload so we can see
+    # what reason (or non-reason) is driving the unregister cascade. Empirically
+    # observed as bursts of 20+ unregister calls in ~30s on long-running
+    # worker sessions; the root cause needs the payload to investigate. Log
+    # is compacted to one line via tr so it doesn't fragment radio.log.
+    if [[ -n "$payload" ]]; then
+      payload_oneline=$(printf '%s' "$payload" | tr '\n' ' ' | tr -s ' ')
+      _log "unregister: proceeding (reason=${reason:-<unset>}) role=$TASK_FORCE_ROLE payload=$payload_oneline"
+    fi
   fi
 
   local f
   f=$(_session_file "$TASK_FORCE_ROLE")
   rm -f "$f"
+  # Sweep the loadout sidecar too — leaving it behind would let a future
+  # _ensure_session_file resurrect a stale loadout for a role whose tab has
+  # really exited and been re-used (#140).
+  rm -f "$(_loadout_sidecar "$TASK_FORCE_ROLE")"
   _log "unregister role=$TASK_FORCE_ROLE"
 }
 
--- a/claude-gh/bin/radio
+++ b/claude-gh/bin/radio
@@ -385,6 +385,30 @@ cmd_register() {
     fi
   fi
 
+  # Persist the loadout sidecar BEFORE the session file (#150 round-2 review:
+  # Finding 1). If the process is killed (OOM, task-done --force, host
+  # shutdown) between these two _atomic_write calls, the order matters:
+  #
+  #   sidecar-first (this order):
+  #     pre-kill state:  .loadout=claude-gh, .info absent
+  #     re-seed reads:   .loadout → LOADOUT=claude-gh recovered ✓
+  #
+  #   info-first (the OLD order — pre-#150-r3):
+  #     pre-kill state:  .info has LOADOUT=claude-gh, .loadout absent
+  #     unregister cascade deletes .info; .loadout never existed
+  #     re-seed reads:   no sidecar, hook subshell has no $TASK_FORCE_LOADOUT
+  #                      → LOADOUT=unknown ✗
+  #
+  # The window is microseconds wide, but the cascade in #140 makes it
+  # statistically reachable. Reversing the order eliminates the failure mode
+  # entirely. Skipping the sidecar on empty $loadout keeps the sidecar's mere
+  # existence equivalent to "we have a real loadout to restore" — otherwise
+  # a re-seed could read an empty sidecar and clobber the env-var fallback
+  # that test paths still rely on.
+  if [[ -n "$loadout" ]]; then
+    printf '%s' "$loadout" | _atomic_write "$(_loadout_sidecar "$role")"
+  fi
+
   {
     printf 'ROLE=%s\nTAB=%s\nTAB_ID=%s\nREPO=%s\nSTATE=idle\nLAST_HEARTBEAT=%s\nAGENT=%s\nLOADOUT=%s\n' \
       "$role" "$tab" "$tab_id" "$repo" "$(_now)" "$agent" "$loadout"
@@ -401,15 +425,6 @@ cmd_register() {
   } | _atomic_write "$f"
 
   mkdir -p "$(_inbox_dir "$role")" "$(_processed_dir "$role")"
-
-  # Persist the loadout name to a sidecar so _ensure_session_file can recover
-  # it during a mid-life re-seed (#140 Bug C). Skipping on empty values keeps
-  # the sidecar's mere existence equivalent to "we have a real loadout to
-  # restore" — otherwise a re-seed could read an empty sidecar and clobber
-  # the env-var fallback that test paths still rely on.
-  if [[ -n "$loadout" ]]; then
-    printf '%s' "$loadout" | _atomic_write "$(_loadout_sidecar "$role")"
-  fi
 
   _log "register role=$role tab=$tab tab_id=$tab_id agent=$agent loadout=$loadout"
 
--- a/claude-jira/bin/radio
+++ b/claude-jira/bin/radio
@@ -200,9 +200,20 @@ _ensure_session_file() {
   local loadout=""
   local sidecar
   sidecar=$(_loadout_sidecar "$TASK_FORCE_ROLE")
-  if [[ -f "$sidecar" ]]; then
-    loadout=$(<"$sidecar")
-  fi
+  # TOCTOU-safe read (#150 review): the sidecar can be unlinked by a
+  # concurrent cmd_unregister between any pre-check and the read — the
+  # unregister cascade documented in #140 is exactly that race window.
+  # `cat … 2>/dev/null || true` collapses three failure modes into "empty
+  # string" without tripping set -e: file absent, file unlinked mid-read,
+  # file unreadable. The env-var fallback below handles the empty case.
+  #
+  # Why not the obvious `loadout=$(<"$sidecar" || true)` the reviewer
+  # suggested: bash's `<file` is a command-substitution-only shorthand, not
+  # a real command, and `<file || true` short-circuits such that the read
+  # never returns the file's contents — so the sidecar value would be lost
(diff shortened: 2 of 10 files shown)


martin-conur:
## PR Review — Radio session corruption fix (#140)

### Spec compliance

The issue documented three interacting bugs (A cascade, B TAB_ID loss, C LOADOUT loss) and proposed explicit fix paths. Checking each:

**Fix A (cascade root cause)** — Instrumentation only, as stated in the PR. `cmd_unregister` now logs the full payload when proceeding past the skip-list. The PR is explicit this is a data-gathering step, not a fix. ✓ matches spec.

**Fix B (TAB_ID lost on re-seed)** — `_zellij_tab_id_by_name` now matches `⏸️ <slug>`, `▶️ <slug>`, and `❓︎ <slug>` in addition to the bare name. The fix is symmetric across `cmd_register` and `_ensure_session_file` as the spec requested. Applied identically across all 7 loadout scripts; drift check confirms byte-for-byte parity. ✓ matches spec.

**Fix C (LOADOUT lost on re-seed)** — Sidecar mechanism (`<role>.loadout`) is implemented: written at register time, read during re-seed, swept on real exit, left intact on skip-list hits (`clear|resume`). However, see Finding 1 below — the sidecar is **not preserved** when the cascade fires with a non-skip reason (e.g. `reason=logout` or `reason=<unset>`), which is the empirically-observed cascade pattern from #140. The sidecar recovery only activates if the session file is deleted by something other than `cmd_unregister`. Partial fix — the full fix requires expanding the skip-list once Fix A's instrumentation identifies the cascade trigger. ⚠️ narrower than the spec implies.

Test coverage is thorough: 12 new bats cases covering the re-seed, sidecar write/read/sweep, skip-list boundary, and payload logging. 770/770 passing.

---

### Code-review findings

**Finding 1 — Bug C not fixed for the documented cascade pattern (design gap)**

The sidecar sweep in `cmd_unregister` (`rm -f "$(_loadout_sidecar "$TASK_FORCE_ROLE")"`, line ~458) runs whenever the skip-list doesn't fire. The #140 log shows the cascade fires with non-skip reasons (empirically `reason=logout`, `reason=<unset>`). Each cascade call deletes both the session file AND the sidecar. When `_ensure_session_file` re-seeds, there is no sidecar to read — it falls back to `${TASK_FORCE_LOADOUT:-unknown}`.

The fix path is correct in principle: once Fix A reveals the cascade trigger reason and it's added to the skip-list, the cascade will preserve the sidecar, and re-seeds will recover LOADOUT correctly. But as-is, the symptom described in the issue (LOADOUT=unknown after cascade) is not resolved for the observed pattern. Worth calling out explicitly so the follow-up issue for Fix A is filed as a blocker for Bug C, not a nice-to-have.

**Finding 2 — `lib/zellij-tab.sh` emoji mismatch (pre-existing, newly exposed)**

`_rename_tab` in `radio` writes the `awaiting` prefix as `U+2753 U+FE0E` + space (❓︎ with VS-15 text-presentation variation selector). `lib/zellij-tab.sh`'s `sub("^❓ "; "")` strip (line ~49) uses `U+2753` + space — no VS-15. These byte sequences don't match. When `aw_zellij_tab_id_by_name` runs against a tab in the `awaiting` state, it fails to strip the prefix and the lookup misses. This was not introduced by this PR (the mismatch predates it) but this PR adds the `❓︎` variant to `radio`'s `_zellij_tab_id_by_name` and is the natural place to also sync `lib/zellij-tab.sh`. The issue spec says the fix should be "symmetric across … `_rename_tab`'s tab-name lookups" — `lib/zellij-tab.sh` is a parallel implementation of the same lookup. There's a comment in the file saying "Keep this prefix list in sync with `_rename_tab` in `radio`" that was missed.

**Finding 3 — TOCTOU between sidecar `-f` guard and read (narrow but real)**

In `_ensure_session_file`:
```bash
if [[ -f "$sidecar" ]]; then
  loadout=$(<"$sidecar")   # ← sidecar can be deleted here by concurrent unregister
fi
```
The script runs with `set -euo pipefail`. If a concurrent `cmd_unregister` deletes the sidecar between the `-f` check and the `$(<)` read, the command substitution exits non-zero and `set -e` aborts `_ensure_session_file`. Session file is never written; STATE tracking stops for the worker's lifetime. Fix is one character: `loadout=$(<"$sidecar" || true)`.

**Finding 4 — jq multi-match regression (rare)**

The new `select()` filter matches both `"worker-foo"` and `"⏸️ worker-foo"` when both are present. `.[0]` returns whichever appears first in zellij's output. If a user has a leftover bare-slug tab alongside the painted worker tab, the wrong tab ID can be returned. The old exact-match never had this ambiguity. Unlikely in practice but a genuine regression from the pre-fix semantics; worth adding a note or a filter that prefers the painted variant.

**Nit — `TASK_FORCE_AGENT` has the same hook-subshell env problem as `LOADOUT` (pre-existing)**

`_ensure_session_file` writes `AGENT=${TASK_FORCE_AGENT:-claude}`. `$TASK_FORCE_AGENT` is not available in the Claude Code hook subshell for the same reason `$TASK_FORCE_LOADOUT` wasn't. For kiro-* workers this means re-seeded sessions write `AGENT=claude` instead of `AGENT=kiro`. Currently AGENT is not consumed by any `cmd_send` path so there is no runtime impact — but if that changes, this becomes a silent bug. Not in scope for this PR (pre-existing) but worth tracking.

---

### Verdict: **clean-with-nits**

The three-bug diagnosis is correct and the fix for Bug B (TAB_ID) is solid. Bug C's sidecar mechanism is well-implemented but only activates when the session file is deleted outside of `cmd_unregister`; for the cascade scenario it waits on Bug A's instrumentation to identify the skip-list addition. The TOCTOU race (Finding 3) is a one-line fix worth patching before merge. The `lib/zellij-tab.sh` emoji sync (Finding 2) is pre-existing but should be filed as a follow-up since the comment explicitly says to keep the prefix list in sync. Findings 1 and 4 are design-level observations rather than blockers.

Suggested before merge:
- [ ] Fix the TOCTOU: `loadout=$(<"$sidecar" || true)` in `_ensure_session_file` (all 7 scripts)
- [ ] File a follow-up issue: "Bug C fix is incomplete for non-skip cascade reasons — blocked on Bug A skip-list expansion"

martin-conur:
## Changes requested — round 2

Reviewer's 4 findings + 1 nit triaged. Two small fixes pre-merge; rest filed as follow-ups.

### Pre-merge

**#3 — TOCTOU on sidecar read** (reviewer's only \"before merge\" ask)

In `_ensure_session_file`:
```bash
if [[ -f "$sidecar" ]]; then
  loadout=$(<"$sidecar")   # ← sidecar can be unlinked here by concurrent unregister
fi
```

`set -e` aborts the function on the read failure; session file never written; STATE tracking dies for the worker's life. One-liner fix:
```diff
- loadout=$(<"$sidecar")
+ loadout=$(<"$sidecar" || true)
```

Apply across all 7 drift-grouped radio scripts. Add a regression bats case that deletes the sidecar between the `-f` check and the read (concurrent unregister simulation) and asserts `_ensure_session_file` still completes.

**#2 — `lib/zellij-tab.sh` emoji-strip out of sync with `radio`'s `_rename_tab`**

`_rename_tab` writes the awaiting prefix as `❓︎ ` (U+2753 + U+FE0E variation selector + space). `lib/zellij-tab.sh`'s strip uses `❓ ` (U+2753 + space, no VS-15). They don't match byte-for-byte → `aw_zellij_tab_id_by_name` fails to strip when a tab is in awaiting state.

This PR adds the `❓︎` variant to `radio`'s lookup, but `lib/zellij-tab.sh:49` (and the corresponding new lookup logic) needs the same fix to stay in sync. The file even has a comment saying so:

> Keep this prefix list in sync with `_rename_tab` in `radio`.

Fix is replacing the strip in `lib/zellij-tab.sh` with the same three-variant strip (`⏸️ ` / `▶️ ` / `❓︎ `), matching `radio` byte-for-byte. Single file, no drift group concerns.

### Filed as follow-ups (do NOT touch in this PR)

- **#151** — Complete #140: analyze cascade-payload logs (this PR's diagnostic adds them), expand `cmd_unregister`'s skip-list to include the observed reasons, restore LOADOUT/AGENT preservation. Reviewer #1 + AGENT nit. P1 / milestone v0.3.0 — second half of #140's fix.
- **#152** — `_zellij_tab_id_by_name` multi-match regression (reviewer #4): jq filter ambiguity when bare-slug + painted variants coexist. P2, no milestone.

### Why we're not doing #1 in this PR

Reviewer correctly diagnoses that completing Bug C requires Fix A's actual root-cause (skip-list expansion), which is gated on diagnostic data this PR adds. Trying to finish it in #150 would mean guessing the cascade reason and shipping a speculative skip-list — wrong shape. Better to: merge #150 (Fix B + Fix C mechanism + Fix A logging), collect real cascade logs from a dogfooded worker session, then close out #151 with the actual observed reasons.

### Out of scope for round 2

Just the two small fixes. Radio `re-review-requested` when ready.

martin-conur:
## PR Review — Radio session corruption round 2 (#140)

### Spec compliance

The round-2 spec called for exactly two pre-merge fixes (TOCTOU sidecar read + lib/zellij-tab.sh emoji sync) and two filed follow-ups (#151, #152). Checking each:

**TOCTOU fix (Finding 3 from round 1)** — The reviewer suggested `loadout=$(<"$sidecar" || true)`; the PR rejects this with a correct explanation: bash's `$(<file)` in a command substitution treats `|| true` as operating on the substitution group exit status, not the read itself, so the file content is lost even on the happy path. The chosen fix (`cat "$sidecar" 2>/dev/null || true`) correctly collapses file-absent, mid-read-unlink, and unreadable into empty string without tripping `set -e`. Applied across all 7 radio scripts. ✓ matches spec.

**lib/zellij-tab.sh emoji sync (Finding 2 from round 1)** — `sub("^❓ "; "")` (U+2753 + space) is updated to `sub("^❓︎ "; "")` (U+2753 + U+FE0E + space), matching `_rename_tab`'s bytes in all 7 radio scripts. New bats test in `claude_gh_task_work.bats` (line 1114) exercises the awaiting-state lookup and passes. ✓ matches spec.

**#151 (Bug C completion + skip-list expansion)** — filed as follow-up, not touched. ✓  
**#152 (multi-match jq regression)** — filed as follow-up, not touched. ✓

Test coverage: 770/770 passing, 12 new cases covering round-2 scenarios. ✓

---

### Code-review findings

**Finding 1 — Sidecar write order creates a narrow LOADOUT-loss window (CONFIRMED)**

In `cmd_register`, the session file (`.info`) is written first via `_atomic_write`, then the loadout sidecar (`.loadout`) is written second:

```bash
} | _atomic_write ""            # .info written first
mkdir -p ...
if [[ -n "" ]]; then
  printf '%s' "" | _atomic_write "")  # .loadout written second
fi
```

If the process is killed (OOM, `task-done --force`, host shutdown) between these two writes, `.info` exists with the correct `LOADOUT=claude-gh` value, but `.loadout` is absent. The session file itself is fine — the problem only surfaces later:

1. A subsequent unregister cascade fires → deletes `.info`; `rm -f` on the absent `.loadout` is a no-op.
2. `_ensure_session_file` re-seeds: no `.info`, no `.loadout`, `reviewer` not in hook subshell env → writes `LOADOUT=unknown`.

The fix is one-line: swap the write order — write the sidecar **before** the session file. If killed between the two writes in the reversed order, `.loadout` exists and `.info` is absent — the next re-seed finds the sidecar and correctly recovers `LOADOUT=claude-gh`. There is no downside to the reversed order: `_ensure_session_file` only runs when `.info` is missing, and `cmd_unregister` sweeps the sidecar regardless of whether `.info` exists.

The window is microseconds-wide, but the unregister cascade documented in #140 makes it statistically reachable (20+ rapid-fire unregister calls that can catch any brief inter-write gap).

**Finding 2 — TOCTOU test comment describes removed code (nit)**

The test `re-seed survives when sidecar is unlinked between [[ -f ]] and read` (line 1284) has a comment that says:

> We can't easily race a real unregister against the read in bats, but we can produce the same effect by replacing the read step with a deletion: a bash DEBUG trap that fires *between* `[[ -f ]]` and the read, deletes the sidecar…

But the PR removed the `[[ -f "$sidecar" ]]` guard entirely — the actual code now does an unconditional `cat "$sidecar" 2>/dev/null || true` with no pre-check. The race window the comment describes no longer exists in the form it portrays. A future reader will search for the `[[ -f ]]` guard referenced and not find it. Additionally, the chmod-0 test never asserts the fallback value (`LOADOUT=unknown`), only `assert_success` and `STATE=busy` — which would pass even if `cat` succeeded. The companion unlink-variant test (line 1324) is correct and asserts `LOADOUT=unknown`; this one should do the same.

**Finding 3 — Multi-match jq `.[0]` ambiguity (PLAUSIBLE, acknowledged as #152)**

Already filed as follow-up #152 and noted in the previous review. Confirming: verified via source inspection that `cmd_send`'s wake-up path (line 553) reads `TAB_ID=` directly from the session file, bypassing `_zellij_tab_id_by_name` entirely — so the ambiguity has no impact on the primary message-delivery path. The lookup is only used during `cmd_register` and `_ensure_session_file`, both of which run before `_rename_tab` has painted the tab, so in normal operation there is only one matching entry. PLAUSIBLE only with manual out-of-band tab creation or a zellij bug. ✓ #152 is the right tracking vehicle.

---

### Verdict: **clean-with-nits**

Round 2 correctly addressed both pre-merge items from round 1. The TOCTOU fix explanation (rejecting the reviewer's suggested `$(<file || true)` form) is accurate and well-documented. The emoji sync in `lib/zellij-tab.sh` is byte-correct and tested.

One new finding worth patching before merge:
- [ ] **Finding 1** — Swap write order in `cmd_register`: write `.loadout` sidecar **before** writing `.info` (applies to all 7 radio scripts). Closes the narrow kill-window where `.info` is written but `.loadout` never gets written.

One test hygiene item:
- [ ] **Finding 2** — Update the TOCTOU test comment (no `[[ -f ]]` guard exists anymore); add `assert_output "LOADOUT=unknown"` to the chmod-0 variant to match the unlink-variant test's assertion.

martin-conur:
## Changes requested — round 3 (both small)

Reviewer's round-2 verdict is clean-with-nits with two contained pre-merge items. Both should land in one push.

### 1. Swap sidecar/info write order in `cmd_register`

In `cmd_register`, `.info` is currently written first, then `.loadout`. If the process dies between the two writes (OOM, `task-done --force`, host shutdown), we get a session file with the correct LOADOUT value but no sidecar — and the next unregister-cascade + re-seed produces `LOADOUT=unknown`.

Fix: reverse the write order — write `.loadout` first, then `.info`. If killed between in the reversed order, the sidecar exists and the re-seed recovers LOADOUT correctly. No downside: `_ensure_session_file` only runs when `.info` is missing; `cmd_unregister` sweeps the sidecar regardless.

Apply across all 7 drift-grouped radio scripts (`radio|body` group). Add a bats case that simulates the kill (write the sidecar, no .info, run `_ensure_session_file`, assert LOADOUT is the sidecar value not `unknown`).

### 2. TOCTOU test hygiene

`tests/radio_lifecycle.bats` line ~1284 — the test `re-seed survives when sidecar is unlinked between [[ -f ]] and read` has a stale comment referencing the `[[ -f "$sidecar" ]]` guard that round 2 removed. The comment describes a race window that no longer exists in the code (we now do unconditional `cat ... 2>/dev/null || true`). Update the comment to describe the actual scenario.

Additionally, the chmod-0 variant doesn't assert `LOADOUT=unknown` — only `assert_success` + `STATE=busy`, which would pass even if `cat` succeeded on a readable file. The unlink variant correctly asserts `LOADOUT=unknown`; the chmod-0 variant should match.

### Out of scope

Don't touch #151 / #152 / anything else. Radio `re-review-requested` when ready.

### Session corruption reminder

Your worker session still has `TAB_ID=""` / `LOADOUT=unknown` from the very bug you're fixing. The radio ping for this round-3 ask will queue but won't auto-wake — user will manually nudge as before. Soon enough this ends.

martin-conur:
## PR Review — Radio session corruption round 3 (#140)

### Spec compliance

Round-3 spec called for exactly two small pre-merge fixes. Checking both:

**1. Sidecar write order in `cmd_register`** (reviewer round-2 Finding 1) — Sidecar (`.loadout`) is now written via `_atomic_write` **before** the session file (`.info`). The pre-kill recovery table in the comment is correct: if the process is killed between the two writes in the new order, the sidecar exists and `_ensure_session_file` can recover LOADOUT; under the old order, the sidecar would never have been written and the re-seed would produce `LOADOUT=unknown`. Applied consistently across all 7 radio scripts. A new bats test (`radio_lifecycle.bats` line ~1491) simulates the mid-kill state (sidecar present, .info absent) and asserts `LOADOUT=claude-gh` is recovered. ✓ matches spec.

**2. Test hygiene** (reviewer round-2 Finding 2) — Two items:
- TOCTOU test comment no longer references the `[[ -f "$sidecar" ]]` guard that round 2 removed. The updated comment accurately describes the unconditional `cat … 2>/dev/null || true` approach and the three failure modes it collapses. ✓
- The chmod-0 variant (line ~1452) now asserts `LOADOUT=unknown`, confirming the env-var fallback fired rather than just that the script completed. Matches the unlink-variant test's assertion. ✓

Test coverage: all 7 radio scripts are byte-identical (confirmed by diff patterns); 770/770 tests passing; 12 new bats cases added across rounds 1–3.

---

### Code-review findings

**Finding 1 — `$(<"$sidecar" || true)` claim verified correct**

The comment defending the choice of `cat` over `$(<"$sidecar" || true)` is accurate. Verified locally:

```bash
bash -c 'echo x>/tmp/f; v=x; echo "[$v]"'
# → []
```

Inside `$()`, `<file` is a redirect for a null command; the null command exits 0, `|| true` never fires, and the subshell produces no output. `cat` is the correct unambiguous form. The explanation in the codebase is accurate and will save the next reader from re-deriving this.

**Nit — stale sidecar after partial unregister, re-registered without `--loadout`**

If `cmd_unregister` is killed between `rm -f "$f"` and `rm -f "$(_loadout_sidecar …)"`, a stale `.loadout` file survives. On the next `cmd_register` for the same role *with* a `--loadout` arg, the sidecar is overwritten immediately (the new write happens before `.info`), so the stale value is harmless. The only remaining gap: if a re-registration *without* `--loadout` fires for the same role, the `[[ -n "$loadout" ]]` guard skips the sidecar write, the stale `.loadout` persists, and a subsequent cascade-triggered re-seed reads the wrong loadout. In practice every `task-work` invocation passes `--loadout`; only manual `radio register` calls without `--loadout` are affected. Low-impact, worth knowing about but not a blocker.

---

### Verdict: **clean**

Both round-3 items are correctly implemented. The write-order explanation is accurate and well-tested. The test comment is updated and the chmod-0 assertion gap is closed. No new correctness issues found in this round.

No pre-merge work required. Following issues remain open for tracking (filed in round 2):
- #151 — Bug C completion + skip-list expansion (blocked on cascade diagnostic data this PR provides)
- #152 — `_zellij_tab_id_by_name` multi-match `.[0]` ambiguity (low-impact, P2)

Issue #152, opened 2026-06-02
_zellij_tab_id_by_name: jq filter can match wrong tab when bare-slug + painted variants coexist


martin-conur:
## Context

Follow-up from PR #150 (closes #140 partially) — reviewer finding #4.

PR #150 changed \`_zellij_tab_id_by_name\` (in \`radio\` and \`lib/zellij-tab.sh\`) from exact-name match to a select() that accepts the bare name OR the painted variants (\`⏸️ <name>\`, \`▶️ <name>\`, \`❓︎ <name>\`). This fixes Bug B (TAB_ID lookup after emoji repaints), but introduces a multi-match ambiguity:

If zellij contains BOTH a bare \`worker-foo\` tab AND a painted \`⏸️ worker-foo\` tab (e.g. a leftover from a previous session or a user-renamed tab), the new jq \`select()\` produces an array of multiple tab IDs. \`.[0]\` returns whichever appears first in zellij's output — unstable across zellij restarts.

The pre-PR exact-name match was unambiguous (an exact-name collision means the user has two tabs with literally identical names, which is rare enough zellij itself surfaces it).

## Practical risk

Low. The bare-slug tab would have to outlive a rename operation, which our radio code only does to its own tabs. Users renaming tabs to literal slugs would be the main cause. Not a security or correctness blocker.

## Proposed fix

Prefer the painted variant when multiple matches exist:

```bash
zellij action list-tabs --json \\
  | jq -r --arg n "$target" '
      def strip: sub("^⏸️ "; "") | sub("^▶️ "; "") | sub("^❓︎ "; "");
      [ .[] | select(strip == $n) ] as $matches
      | (
          ($matches[] | select(.name != $n))    # painted variants first
          // $matches[0]                         # else bare match
        ) | .tab_id // empty'
```

Or simpler: prefer entries whose \`.name\` starts with one of the known prefixes; fall back to exact-bare-name.

## Files

- \`claude-gh/bin/radio\` (\`_zellij_tab_id_by_name\`, drift-grouped via \`radio|body\` — fixes propagate to 6 siblings)
- \`lib/zellij-tab.sh\` (\`aw_zellij_tab_id_by_name\`)
- \`tests/radio_zellij.bats\` — add a multi-match scenario

## Priority: P2

Theoretical regression, low practical likelihood. Worth fixing for correctness but not v0.3.0-blocking.

## References

- PR #150 reviewer finding #4

Issue #151, opened 2026-06-02
Complete #140: analyze cascade-payload logs, expand SessionEnd skip-list, restore LOADOUT/AGENT preservation


martin-conur:
## Context

Follow-up from PR #150 (closes #140 partially) — reviewer findings #1 + nit.

PR #150 ships three pieces for the radio session corruption bug:
- **Fix B** (TAB_ID emoji-aware lookup) — complete.
- **Fix C** (LOADOUT sidecar) — mechanism in place but does **not** activate for the cascade pattern documented in #140 (non-skip reasons like \`reason=logout\` / \`reason=<unset>\`).
- **Fix A diagnostic** — \`cmd_unregister\` now logs the full SessionEnd hook payload when it proceeds past the skip-list.

This issue is the cleanup pass to finish Bug A and unblock Bug C.

## The dependency

PR #150's sidecar is swept whenever \`cmd_unregister\` doesn't bail on the skip-list (\`reason=clear|resume\` only). The cascade fires with non-skip reasons, blasts both the session file AND the sidecar, and \`_ensure_session_file\` then re-seeds with no sidecar to read → \`LOADOUT=unknown\` returns.

The complete fix is:
1. Collect cascade logs from a few real worker runs (PR #150 adds the logging mechanism).
2. Identify which \`reason\` values the cascade actually fires with.
3. Expand \`cmd_unregister\`'s skip-list to include those values (treat the cascade-trigger reasons the same as \`clear\` / \`resume\` — known intra-session events that must not delete state).
4. Re-test the corruption scenario: with the cascade skipped, the sidecar survives, \`_ensure_session_file\` reads it, LOADOUT is preserved.

## Includes: AGENT env-loss (reviewer nit)

\`_ensure_session_file\` writes \`AGENT=\${TASK_FORCE_AGENT:-claude}\`. \`\$TASK_FORCE_AGENT\` has the same hook-subshell unset problem \`\$TASK_FORCE_LOADOUT\` had — kiro-* workers re-seed with \`AGENT=claude\` instead of \`AGENT=kiro\`. No runtime consumer today, but a silent footgun if anyone adds one.

Fix: apply the same sidecar pattern (or whatever mechanism solves LOADOUT) to AGENT. Same place, same code, trivial.

## Deliverables

[...]

Issue #149, opened 2026-06-02
task-done: parse .info file via awk instead of source (defends against opaque non-gh identifiers)


martin-conur:
## Context

Follow-up from PR #145 (closes #144) — reviewer's clean-with-nits finding.

After #144 lands, non-gh loadouts (claude-jira / claude-notion / claude-local) can pass any opaque string as the spec identifier. `task-reviewer` writes it unquoted into the worktree's `.info` file:

```bash
printf 'ISSUE_NUMBER=%s\n' "$ISSUE_NUMBER"
```

Producing lines like:
```
ISSUE_NUMBER=PROJ-$(date)
```

`task-done` then reads the file via `source "$INFO_FILE"` (`claude-gh/bin/task-done:54`). The shell re-parses the value — `$(...)` executes, `$VAR` expands, backticks execute.

## Practical risk: zero

- Real Jira keys: `PROJ-123` — no metacharacters.
- Real Notion URLs: `https://notion.so/...?p=...` — `%`-encoded special chars.
- Real local slugs: `kebab-case-string` — no metacharacters.

The vector requires explicitly crafted identifiers and would be a self-inflicted local operation. Filed for completeness and code-hygiene, not because exploitation is realistic.

## Why fix anyway

- `source` of user-influenced data is a code-smell that ages badly — future contributors adding fields to `.info` may not realize the value gets `eval`'d.
- The fix is small and matches an existing pattern in the same file.
- After #140 (radio session corruption) and #148 (task-done branch leak), `task-done`'s cleanup path is getting attention anyway — bundling the parse-hardening keeps `task-done` polished.

## Proposed fix

Two options:

**Option A — quote on write (defensive at source)**:
```diff
- printf 'ISSUE_NUMBER=%s\n' "$ISSUE_NUMBER"
+ printf 'ISSUE_NUMBER=%q\n' "$ISSUE_NUMBER"
```
…in `task-reviewer`. `task-done`'s `source` then sees a quoted assignment that bash parses safely.

**Option B — replace `source` with awk parser in `task-done`** (defensive at read):

`task-done` already uses an awk-based read for `TAB_ID` (`claude-gh/bin/task-done:116, 121`):
```bash
_WORKER_TAB_ID=$(awk -F= '/^TAB_ID=/ { print $2; exit }' "$INFO_FILE")
```

Extend the same pattern to every field `task-done` currently relies on from `$INFO_FILE`:
- `BASE_BRANCH`
- `SLUG`
[...]

PR #153, opened 2026-06-02
task-done --remove-worktree: force-delete reviewer branch on PR_NUMBER (#148)

### LINKED ISSUE #148: task-done --remove-worktree leaks reviewer branch + state; blocks re-dispatch of task-reviewer
## Symptom

After running `task-done --remove-worktree` in a reviewer's worktree (the documented cleanup path for the dispatch-style reviewer introduced in #138), `task/review-pr<N>` lingers as a local branch. The next `task-reviewer <N>` then refuses to dispatch because its per-PR guard (`claude-gh/bin/task-reviewer:185`) checks both worktree dir AND branch existence:

```
Error: a review worktree or branch already exists for PR #N:
       branch:       task/review-prN
```

The user has to manually `git branch -D task/review-pr<N>` from outside before the re-dispatch will succeed. This hits every re-review round-trip — confirmed twice in dogfooding during PR #142 and #145 review cycles within a single PM session.

## Root cause

`claude-gh/bin/task-done:150` (and the byte-identical bodies across the 4 other claude-* / kiro-* variants via the `task-done-std` and `task-done-local` drift groups) does:

```bash
if git branch -d "$BRANCH" 2>/dev/null; then
  echo "Done. Deleted local branch '$BRANCH'."
else
  echo "Done. Branch '$BRANCH' still has unmerged commits — delete manually after PR merge: git branch -D $BRANCH"
fi
```

`git branch -d` is the safe-delete variant — succeeds only if the branch is fully merged into HEAD (main). For **worker** branches this is correct: you don't want to lose unmerged work.

For **reviewer** branches (`task/review-pr<N>`), the safety is misplaced:
- They're forked from `FETCH_HEAD` (the PR's head ref).
- They contain zero commits beyond the PR's head — they're pure scaffolding for the worktree checkout.
- They are never going to be merged into main directly (the PR they reviewed is what gets merged).
- So `git branch -d` always fails on them while a review is happening on an unmerged PR — exactly when this cleanup runs.

The fall-through "delete manually after PR merge" instruction is misleading: for reviewer branches the merge will never include this branch.

## Discriminator

`task-reviewer` writes a unique key into its `$INFO_FILE`:

```bash
# claude-gh/bin/task-reviewer:233
printf 'PR_NUMBER=%s\n' "$PR_NUMBER"
printf 'ISSUE_NUMBER=%s\n' "$ISSUE_NUMBER"
```

`task-work` does **not** write `PR_NUMBER` to its info file (it writes `BASE_BRANCH`, `SLUG`, `GH_URL`, `TAB_ID` only).

`PR_NUMBER` being set in the sourced info file is the clean signal "this is a reviewer worktree." `task-done` can branch on that.

### Part A — task-done branch cleanup branches on info-file marker

In the `# region:confirm-and-cleanup` block of `claude-gh/bin/task-done` (and the rest of the `task-done-std` drift group), replace:

```bash
if git branch -d "$BRANCH" 2>/dev/null; then
  echo "Done. Deleted local branch '$BRANCH'."
else
  echo "Done. Branch '$BRANCH' still has unmerged commits — delete manually after PR merge: git branch -D $BRANCH"
fi
```

with something like:

```bash
# Reviewer worktrees write PR_NUMBER into their info file; they're pure
# scaffolding and their branch never merges into main (the PR does).
# Force-delete in that case; safe-delete otherwise.
if [[ -n "${PR_NUMBER:-}" ]]; then
  git branch -D "$BRANCH" 2>/dev/null \
    && echo "Done. Deleted reviewer branch '$BRANCH' (-D, scaffold-only)." \
    || echo "Done. Branch '$BRANCH' could not be deleted — clean up manually."
else
  if git branch -d "$BRANCH" 2>/dev/null; then
    echo "Done. Deleted local branch '$BRANCH'."
  else
    echo "Done. Branch '$BRANCH' still has unmerged commits — delete manually after PR merge: git branch -D $BRANCH"
  fi
fi
```

Same change for `task-done-local` group (`claude-local/bin/task-done`, `kiro-local/bin/task-done`).

`PR_NUMBER` is already in scope at that point: the `# region:worktree-context` block sources `$INFO_FILE` (`claude-gh/bin/task-done:54`).

### Part B — verify other reviewer state actually does get cleaned up

Audit and confirm `task-done --remove-worktree` against a reviewer worktree cleans up:

| Artifact | Currently cleaned by task-done? |
|---|---|
| `~/.task-force/radio/sessions/reviewer-*.info` | ✅ via `radio unregister` |
| `<worktrees>/.review-pr<N>.info` | ✅ via `rm -f "$INFO_FILE"` |
| `<worktrees>/review-pr<N>/` (worktree dir) | ✅ via `git worktree remove` |
| `task/review-pr<N>` (local branch) | ❌ this issue's fix |
| Zellij tab | ✅ via `zellij action close-tab-by-id` (#107 fix) |

If audit surfaces additional leaks, fold them in.

### Part C — tests

Extend `tests/task_done.bats` (or `tests/task_done_close_tab.bats`):

- New: invoke `task-done --remove-worktree --force` in a fake reviewer worktree (one with `PR_NUMBER` in its `$INFO_FILE`). Assert `git branch -D` was called instead of `git branch -d`. Assert the branch is gone after.
- Regression: invoke against a non-reviewer (worker) worktree, with `BASE_BRANCH` set but no `PR_NUMBER`. Assert behavior is unchanged — `git branch -d` is tried, falls through to the manual-cleanup message if unmerged.

### Part D — CHANGELOG

`CHANGELOG.md` `[Unreleased]` — `Fixed` entry referencing this issue. Lands before #143's v0.3.0 cut so the release notes mention the dispatch-style reviewer re-review cycle is hands-off.

## Files (expected touchpoints)

| File | Action |
|---|---|
| `claude-gh/bin/task-done` | MODIFY — confirm-and-cleanup region; PR_NUMBER branch |
| `claude-jira/bin/task-done` | MODIFY — same (drift group) |
| `claude-notion/bin/task-done` | MODIFY — same |
| `claude-local/bin/task-done` | MODIFY — same |
| `kiro-gh/bin/task-done` | MODIFY — same (drift group `task-done-std`) |
| `kiro-notion/bin/task-done` | MODIFY — same |
| `kiro-local/bin/task-done` | MODIFY — `task-done-local` group |
| `tests/task_done.bats` | EXTEND |
| `CHANGELOG.md` | ADD Fixed entry |

`tools/check-drift.sh` should stay green — the new branch is inside the existing `confirm-and-cleanup` region.

## Verification

1. `bats tests/task_done.bats` — green.
2. `tools/check-drift.sh` — green (no new groups; existing `confirm-and-cleanup` regions stay byte-identical across the loadout groups they already belong to).
3. End-to-end: PM dispatches `task-reviewer 42` → reviewer comments → user `task-done --remove-worktree` in the reviewer tab → PM dispatches `task-reviewer 42` AGAIN → second dispatch succeeds with no `Error: a review worktree or branch already exists` and no manual `git branch -D` step.
4. Regression: worker dispatches via `task-work` and runs `task-done --remove-worktree` after PR is merged. Branch is still safely deleted via `-d`; if unmerged, the manual-cleanup message still surfaces.

## Why P1 / v0.3.0 blocker

- Bites every reviewer re-review cycle. Each round-trip needs a manual `git branch -D`. We hit it twice in a single PM session today (PR #142 review round 2, PR #145 review round 2).
- The dispatch-style reviewer is one of v0.3.0's headline changes (#138). Shipping it with this cleanup hole means dogfooders run into the wall on their first re-review.
- The fix is small and contained — one region in `task-done`, gated on a marker that already exists in the info file.

## Out of scope

- The "task-done knows what kind of worktree it's in" idea generalized further (e.g. tag worktree type explicitly via a `WORKTREE_KIND=` info field). `PR_NUMBER`-presence is sufficient and zero-migration.
- Auto-running `task-done --remove-worktree` from inside the reviewer tab when the reviewer agent finishes — the #138 design deliberately leaves the tab open so the user can scroll the analysis. Don't change that contract here.
- Backfilling `git branch -D` against already-orphaned reviewer branches in dogfoot repos. Users can `git branch -D task/review-pr*` themselves.

## References

- `claude-gh/bin/task-done:150` — the misplaced `-d` safety check
- `claude-gh/bin/task-reviewer:185` — the per-PR guard that this leak collides with
- `claude-gh/bin/task-reviewer:233` — the `PR_NUMBER` info-file write that's the discriminator
- `tools/check-drift.sh:14-21` — `task-done-std` / `task-done-local` drift groups
- #138 — dispatch-style task-reviewer redesign (the feature this bug paper-cuts)
- #143 — v0.3.0 release cut (this is the third blocker candidate alongside #140 / #144)

martin-conur:
## Summary

- Branch task-done's branch-deletion on the `PR_NUMBER=` marker that `task-reviewer` writes into the worktree's `.info` file. Reviewer worktrees get `git branch -D` (their `task/review-pr<N>` branch is pure scaffolding forked from the PR's head ref and is never going to land in main); worker worktrees keep the safe-delete `git branch -d`.
- Fixes the per-PR guard collision in `task-reviewer` (`Error: a review worktree or branch already exists for PR #N`) that bit every re-review round-trip and forced a manual `git branch -D task/review-pr<N>` between cycles.
- Applied across the `task-done-std` and `task-done-local` drift groups (7 binaries, byte-identical). `tools/check-drift.sh` stays green — the new conditional lives inside the existing `confirm-and-cleanup` region.

Closes #148

## Test plan

- [x] `./run_tests.sh task_done` — 60 tests pass, including 7 new reviewer-branch force-delete tests (one per variant) plus a worker regression for claude-gh.
- [x] `./run_tests.sh drift_check` — 8 tests pass; `task-done-std` and `task-done-local` `confirm-and-cleanup` regions remain byte-identical.
- [x] `tools/check-drift.sh` — green (21 groups checked).
- [x] `./run_tests.sh` — full suite green (783 tests).
- [ ] End-to-end: dispatch `task-reviewer 42` → `task-done --remove-worktree` → re-dispatch `task-reviewer 42` succeeds without manual `git branch -D`.


martin-conur:
## PR Review — #153: `task-done --remove-worktree`: force-delete reviewer branch on PR_NUMBER (#148)

### Spec compliance

Spec issue #148 defined four deliverables. All are present:

| Deliverable | Status |
|---|---|
| **Part A** — Branch deletion in all 7 `task-done` variants conditioned on `PR_NUMBER` marker | ✅ |
| **Part B** — Audit of reviewer-state cleanup (no new leaks found) | ✅ (audit confirmed in PR description; table in spec was already green except for the branch) |
| **Part C** — Tests: reviewer force-delete (7 variants) + worker safe-delete regression | ✅ |
| **Part D** — CHANGELOG `[Unreleased]` Fixed entry | ✅ |

The change is applied consistently across the `task-done-std` (5 binaries: claude-gh, claude-jira, claude-notion, kiro-gh, kiro-notion) and `task-done-local` (2 binaries: claude-local, kiro-local) drift groups, byte-identical within each group as required.

---

### Code-review findings

#### Blocker — `PR_NUMBER` not zeroed before `source "$INFO_FILE"`

**File:** `claude-gh/bin/task-done` (and all 6 sibling variants) — sourcing region (~line 53)

The pre-source initialization pattern already exists for `BASE_BRANCH`:

```bash
BASE_BRANCH="main"
[[ -f "$INFO_FILE" ]] && source "$INFO_FILE"
```

`PR_NUMBER` is **not** given the same treatment. A worker's `.info` file contains no `PR_NUMBER=` line, so sourcing it never writes that variable. If `PR_NUMBER` is set in the ambient shell environment (e.g., the user ran a `task-reviewer` session earlier in the same shell session, or has it exported), `[[ -n "${PR_NUMBER:-}" ]]` will fire on the worker worktree and call `git branch -D` on a branch that may contain unmerged commits — silently destroying work that was deliberately protected by `git branch -d`.

**Fix**: add `PR_NUMBER=` on the line before `source "$INFO_FILE"`, mirroring `BASE_BRANCH="main"`:

```bash
BASE_BRANCH="main"
PR_NUMBER=
[[ -f "$INFO_FILE" ]] && source "$INFO_FILE"
```

Apply to all 7 variants.

**Test gap**: The regression test `claude-gh: worker worktree (no PR_NUMBER) still uses safe-delete -d` does not set `PR_NUMBER=42` in the test environment before invoking `task-done`, so it passes even with the bug present and would not catch this scenario. A follow-on test that exports `PR_NUMBER=42` and then runs `task-done` from a worker worktree (no `PR_NUMBER` in its info file) would close this.

---

#### Nit — Misleading message when reviewer branch is already gone

**File:** `claude-gh/bin/task-done` (and all 6 siblings) — new else branch in force-delete path

```bash
echo "Done. Branch '$BRANCH' could not be deleted — clean up manually: git branch -D $BRANCH"
```

If `git branch -D` fails because the branch was already deleted (double-invocation, or the user ran `git branch -D` manually before `task-done`), this message instructs the user to run the exact command that already succeeded. The script exits 0 and the worktree is gone, so there's nothing to actually clean up. Not a blocker — the flow still exits cleanly — but worth noting if message accuracy matters for the dogfooding experience.

---

#### Minor — Worker regression test covers only `claude-gh`

**File:** `tests/task_done.bats`

The spec asked for a regression test against a non-reviewer worktree to confirm `git branch -d` behavior is unchanged. The test exists but only for `claude-gh`. The other 6 variants have the reviewer force-delete test but no corresponding worker-path assertion. This is a coverage gap; not a blocker given the drift-check suite, but a future copy-paste in one of the sibling `task-work` scripts that accidentally introduces `PR_NUMBER=` into a worker's info file would not be caught by the sibling's test.

---

### Verdict: `changes-requested`

One blocker: `PR_NUMBER` must be zeroed before `source "$INFO_FILE"` in all 7 variants so an ambient shell export cannot trigger `git branch -D` on a worker branch with unmerged commits. The rest of the implementation (logic, drift consistency, CHANGELOG, test structure) is solid. The nits are non-blocking.

martin-conur:
## Changes requested — one real blocker (data-loss risk)

### Blocker — `PR_NUMBER` not zeroed before `source "$INFO_FILE"`

`BASE_BRANCH="main"` is correctly defaulted before the source (`task-done:52`); `PR_NUMBER` is not. Worker `.info` files contain no `PR_NUMBER=` line, so sourcing one never touches that variable. If `PR_NUMBER` is set in the ambient shell (e.g., the user ran `task-reviewer 42` earlier in the same shell session, leaving the export in place, then `cd`'d into a worker worktree and ran `task-done`), the new `[[ -n "${PR_NUMBER:-}" ]]` guard fires on a worker branch and calls `git branch -D` — silently force-deleting any unmerged worker commits.

**Fix** — mirror the existing default. In all 7 task-done variants (`task-done-std` + `task-done-local` drift groups):

```diff
  BASE_BRANCH="main"
+ PR_NUMBER=
  [[ -f "$INFO_FILE" ]] && source "$INFO_FILE"
```

**Test gap** — the existing regression test `worker worktree (no PR_NUMBER) still uses safe-delete -d` does NOT export `PR_NUMBER=42` in the test environment before invoking `task-done`. It only proves the conditional doesn't fire when both file and environment are absent — i.e., it would silently pass even with the bug present. Add a new test case:

```bash
@test "ambient PR_NUMBER export does NOT trigger force-delete on worker" {
  # Ambient export — simulates user having run task-reviewer earlier this shell.
  export PR_NUMBER=42
  # Build a worker worktree info file (no PR_NUMBER inside).
  # Run task-done.
  # Assert: git branch -d was called (safe delete), NOT -D, and the worker branch's
  # unmerged commit is preserved.
  unset PR_NUMBER
}
```

This is the single test that would catch the bug. Without it, future maintainers could revert the `PR_NUMBER=` zero-init and CI stays green.

### Non-blocking nits (take while you're in there or skip)

**Misleading message when branch already gone**: when `git branch -D` fails because the branch was already deleted (double-invoke, or user ran it manually before `task-done`), the error message instructs the user to run the exact command that already succeeded. Worth a graceful "already deleted" branch — but the flow exits cleanly so it's not load-bearing.

**Worker regression coverage only on claude-gh**: the worker-path safe-delete assertion runs only against claude-gh's task-done. Other 6 variants get the reviewer force-delete test but no worker-path assertion. Drift-check covers structural copy, but if a future sibling change drifts the worker-path behavior, only claude-gh's test catches it. Low priority since the bodies are byte-identical inside the drift group regions.

### Out of scope

Just the PR_NUMBER zero-init + the ambient-export regression test. Don't touch #151 / #152 / #149.

### Note on worker session

Your TAB_ID is populated (post-#150 fix is active), so this ping should auto-wake you. LOADOUT=unknown is expected until #151 lands — doesn't affect radio routing.

martin-conur:
## Re-review — PR #153 (round 2)

### Context

This is a re-review after the changes-requested findings from round 1. The prior blocker was: `PR_NUMBER` not zeroed before `source "$INFO_FILE"`, allowing an ambient shell export to trigger `git branch -D` on a worker branch with unmerged commits. Round 1 also asked for an ambient-export regression test to prove that the zero-init actually blocks the leak.

---

### Changes-requested blocker: resolved ✅

All 7 `task-done` variants now have:

```bash
PR_NUMBER=
[[ -f "$INFO_FILE" ]] && source "$INFO_FILE"
```

This is byte-identical to the `BASE_BRANCH="main"` pattern that precedes it — consistent with the existing convention, and verified present in every file in the drift groups.

The ambient-export regression test (`assert_ambient_pr_number_does_not_force_delete`) is now present, exercising:
- `claude-gh` (representative of the 5-file `task-done-std` drift group)
- `claude-local` (representative of the 2-file `task-done-local` drift group)

The test correctly exports `PR_NUMBER=42` before invoking the script via `bash -c` (child process inherits the export), confirms `git branch -d` is used (not `-D`), and asserts the unmerged commit SHA survives. Mechanically sound.

---

### Code-review findings (this round)

#### Nit — ambient-export regression test covers 2 of 7 variants

`tests/task_done.bats` calls `assert_ambient_pr_number_does_not_force_delete` for `$CLAUDE_GH_TASK_DONE` and `$CLAUDE_LOCAL_TASK_DONE` only. The other 5 variants (`claude-jira`, `claude-notion`, `kiro-gh`, `kiro-notion`, `kiro-local`) have no ambient-export test.

**Mitigating factor**: `tools/check-drift.sh` enforces byte-identical `worktree-context` regions across both drift groups — `task-done-std` (5 files) and `task-done-local` (2 files). The drift check fails if any variant's region diverges. The test comment even names this rationale explicitly. So the 2-of-7 coverage is a deliberate and documented tradeoff, not an oversight. Non-blocking.

**Residual risk**: if a future change splits a variant out of its drift group and does not add a dedicated ambient test, the regression would go undetected for that variant. Worth noting as a follow-on if the drift groups are ever refactored.

#### Nit — worker safe-delete regression covers only `claude-gh`

`"claude-gh: worker worktree (no PR_NUMBER) still uses safe-delete -d"` is the only test asserting that a worker branch with unmerged commits is NOT force-deleted when `PR_NUMBER` is absent. The other 6 variants have the reviewer force-delete test but no worker-path assertion. Same drift-check mitigation applies; non-blocking.

#### Nit — worker regression uses `--force` instead of `--remove-worktree --force`

The reviewer force-delete tests use `--remove-worktree --force`; the worker regression test uses `run_task_done "$CLAUDE_GH_TASK_DONE" --force` (no `--remove-worktree`). These can traverse different code paths before reaching the branch-deletion guard. A future change that accidentally gates the `PR_NUMBER` guard on `REMOVE_ONLY` would pass both the reviewer test and this worker test but silently force-delete worker branches when `task-done --remove-worktree` is called without `PR_NUMBER`. Non-blocking; easy to address if the regression test is ever extended.

#### Nit (pre-existing, noted in round 1) — misleading message when reviewer branch is already gone

When `git branch -D "$BRANCH"` fails because the branch was already deleted (double-invoke), the error message instructs the user to run the exact command that already succeeded or that they already ran. Exit code is 0 and cleanup is otherwise complete. No functional impact; message accuracy only.

---

### TOCTOU candidate (evaluated, refuted)

One angle raised a concern that `rm -f "$INFO_FILE"` at line 153 precedes the branch-delete guard at line 162, creating a crash-between window where a retry would have no info file and therefore an empty `PR_NUMBER`. After reading the full script: the script calls `cd "$MAIN_WORKTREE"` before the info-file deletion, and a re-entry is blocked immediately by the worktree guard (`"Error: you're in the main repo, not a worktree"`). The retry cannot reach the branch-delete code through any normal path. **Not a bug.**

---

### Spec compliance (final check)

| Deliverable | Status |
|---|---|
| Part A — `PR_NUMBER` zero-init + force-delete in all 7 variants | ✅ |
| Part B — audit confirms no new leaks beyond the branch | ✅ |
| Part C — reviewer force-delete tests (7) + worker regression + ambient-export tests (2) | ✅ |
| Part D — CHANGELOG `[Unreleased]` Fixed entry | ✅ |

`tools/check-drift.sh` should stay green — both the `worktree-context` and `confirm-and-cleanup` regions are byte-identical within each drift group.

---

### Verdict: `clean-with-nits`

The round-1 blocker is resolved. The `PR_NUMBER=` zero-init is in place across all 7 variants, the ambient-export regression test covers one file per drift group and the test logic is sound. Remaining findings are all nits — coverage gaps documented and mitigated by drift enforcement, not holes. Ready to merge at PM's discretion.

Issue #168, opened 2026-07-05
radio register: drain-on-register — inject offline backlog summary at SessionStart


martin-conur:
## Problem

Messages sent while a role is offline (e.g. worker reports sent when no PM is running — 35 observed `no session for pm` queues) sit in the inbox when the role finally starts. Nothing at startup surfaces them; the agent begins its session blind to its backlog.

## Design

Drain-on-register: at the end of `cmd_register` (and `_ensure_session_file` re-seed), if the role's inbox is non-empty, print an injected-context summary. `SessionStart` hook stdout **is added to the model's context**, so simply printing is sufficient — no zellij wake needed for one's own session:

```
[radio] N message(s) were queued while this role was offline: <id> from=… intent=… | … Run `radio check` and process them before taking new work.
```

Keep it on stdout only when `$TASK_FORCE_ROLE` is set (silent for plain `claude`, #93). This is the natural companion to the stop-hook flush (busy case) and prompt-hook injection (idle case): register covers the offline→online transition, closing the last delivery gap.

## Files

- `claude-*/bin/radio` (`cmd_register`, `_ensure_session_file`)
- Tests: extend `tests/radio_lifecycle.bats`

## Acceptance criteria

- Start `task-pm` with 3 queued reports → PM's first turn opens with the backlog summary in context.
- Fresh register with empty inbox → no extra stdout.
- Re-seed path (`_ensure_session_file`) does not spam: only emit on actual register (SessionStart), not on every busy/ready self-heal.

martin-conur:
Implementation notes post-#163 (PR #172): implement AFTER #164 and reuse its `_hook_payload_field` / `_inbox_summary` helpers — with those in place this is small: at the end of cmd_register (real register only, NOT the _ensure_session_file re-seed path — spec's no-spam criterion), print the `_inbox_summary` backlog block to stdout; SessionStart stdout is injected into context. Empty inbox → zero stdout. Silent no-role exit per #93. No installer/settings changes needed (SessionStart already runs `radio register`), so no migration this time — just bin/radio + tests. Covers the offline→online gap: the 35 'no session for pm' strandings from the July investigation get surfaced the moment a PM starts.

Issue #164, opened 2026-07-05
radio prompt-hook: surface pending inbox via UserPromptSubmit context injection


martin-conur:
## Problem

An **idle** agent fires no hooks, so if the send-time zellij wake fails (no pane, stale TAB_ID, session file missing) the queued message is invisible until a later wake happens to succeed. The failure mode lands precisely on recipients that are ready to work. Live log shows 20 `no writable pane` + 12 `no TAB` occurrences — each one a message that dangled.

## Design

UserPromptSubmit hook stdout **is injected into the model's context** (unlike Stop). Add `radio prompt-hook` and switch the UserPromptSubmit hook from `radio busy` to it:

1. `_update_state busy` (existing behavior preserved).
2. If the role's inbox is non-empty, print a compact summary to stdout:
   ```
   [radio] 2 unread message(s): <id> from=pm intent=changes-requested pr=41 | <id> from=pm intent=approved-and-merged pr=43. Process with `radio check` / `radio read <id>` before or after the user's request, as appropriate.
   ```
3. Empty inbox → print nothing.

Result: any human interaction with an idle agent surfaces its backlog immediately, even when every wake failed. Combined with the `radio stop-hook` ticket, the mailbox becomes eventually-consistent: hooks are the reliable path, `write-chars` wake is just the latency optimization.

## Files

- `claude-*/bin/radio`
- `.claude/settings.json` template per claude loadout installer + this repo's `.claude/settings.json`
- Tests: `tests/radio_prompt_hook.bats`

## Acceptance criteria

- Prompting an idle worker with 2 queued messages → both listed in injected context; agent can act on them.
- Empty inbox → zero stdout (no context noise).
- No `$TASK_FORCE_ROLE` → silent exit 0 (must not block UserPromptSubmit, #93 semantics).

martin-conur:
Implementation notes post-#163 (PR #172), for whoever picks this up:

- **Reuse the #163 patterns**: the task-init jq migration now has an `upgrade_stop` precedent — this ticket needs the analogous in-place upgrade of the legacy `radio busy` UserPromptSubmit command to `radio prompt-hook`, in all four claude-* installers, plus the post-write stray-variant warning (#172 round-1 finding 5 established that contract). Note `radio busy` is ALSO the PostToolUse command — upgrade only the UserPromptSubmit entry, leave PostToolUse alone.
[...]

Issue #146, opened 2026-06-01
kiro-gh/bin/task-reviewer: migrate to SPEC_IDENTIFIER + AW_LOADOUT to match claude-* (post-#144)


martin-conur:
## Context

Follow-up from PR #145 (closes #144) — reviewer finding #2.

The four `claude-*/bin/task-reviewer` dispatchers were migrated to detect `$AW_LOADOUT` from `${BASH_SOURCE[0]}` and accept any opaque `SPEC_IDENTIFIER` for non-gh trackers. `kiro-gh/bin/task-reviewer` was left untouched in #145 — explicitly out of scope per the spec, since kiro-gh is GitHub-only today and no functional regression.

## Why this still matters

- Inconsistent UX. A user running `task-reviewer 42 PROJ-123` against a kiro-gh project gets `"could not parse issue number"`, while the same invocation against a claude-* project (after #144 lands) accepts the identifier.
- Blocks future `kiro-jira` / `kiro-notion` / `kiro-local` task-reviewer extensions — those would need the same dispatcher logic.
- Drift: `kiro-gh/bin/task-reviewer:145` still has the old `ISSUE_URL=""` + GitHub-only `parse_issue_number`, while its claude-gh sibling now has the new shape.

## Scope

Mirror PR #145's Part A changes into `kiro-gh/bin/task-reviewer`:
- Detect \`\$AW_LOADOUT\` from \`\${BASH_SOURCE[0]}\` (already kiro-gh, but use the same pattern for forward-compat with kiro-jira/notion/local).
- Switch to \`SPEC_IDENTIFIER\` model; for kiro-gh, keep current GitHub-numeric/URL parsing + PR-body \`Closes #N\` auto-detect (same as claude-gh path).
- No changes needed to kiro-gh's reviewer agent — kiro doesn't have \`commands/*.md\` slash commands; its reviewer agent definition lives in \`kiro-gh/agents/reviewer.json\` and is already GitHub-flavored, which is fine for kiro-gh-only.

## Out of scope

- Extending task-reviewer to kiro-jira / kiro-notion / kiro-local — those loadouts don't have task-reviewer at all yet (per #138 scope). Track separately.
- Kiro agent JSON updates beyond what's needed to accept the identifier.

## Verification

- \`tools/check-drift.sh\` — green.
- \`tests/kiro_*\` — extend as needed (mirror the equivalent claude-gh assertion in \`tests/task_reviewer.bats\`).
[...]

PR #172, opened 2026-07-06
radio stop-hook: flush queued messages when the agent goes idle (#163)

### LINKED ISSUE #163: radio stop-hook: flush queued messages when the agent goes idle (Stop-hook block JSON)
## Problem

`radio send` makes exactly one wake attempt, at send time. If the recipient is `busy`/`awaiting` (the common case for a working agent), the message queues with **zero redelivery**. `cmd_ready`'s usage string promises "mark this role idle **(and process pending)**" but the implementation only flips STATE. The Stop hook (`radio ready && radio check`) is useless for delivery: Stop-hook stdout goes to the hook subshell, never to the model.

**Evidence** (live `~/.task-force/radio/log`): 101 `busy … queued` entries with no redelivery; 23 messages currently dangling in the pm inbox (June 24 – Jul 1), all `re-review-requested` / `review-complete-*` reports the PM never saw.

## Design

Add a `radio stop-hook` subcommand and switch the Stop hook to it (replacing `radio ready && radio check`):

1. Read the Stop-hook JSON payload from stdin.
2. If `stop_hook_active` is `true` → just `_update_state idle`, exit 0 (prevents block loops).
3. `_update_state idle`.
4. If the role's inbox has ≥1 message: emit
   ```json
   {"decision": "block", "reason": "radio: N unread message(s) — run `radio check`, process each with `radio read <id>`, act on the intents, then stop."}
   ```
   on stdout and exit 0. Claude Code then forces the agent to continue and drain the queue the moment its turn ends.
5. Empty inbox → plain exit 0 (normal stop).

Keep `radio ready` as-is for manual use, but fix its usage string (drop the false "(and process pending)") or make it also print a pending-count hint.

## Files

- `claude-*/bin/radio` (all claude loadouts; `lib`/drift-checked copies)
- Hook wiring: `.claude/settings.json` template written by `task-init claude-*` (each loadout's installer) + this repo's own `.claude/settings.json`
- Tests: extend `tests/radio_lifecycle.bats` / new `tests/radio_stop_hook.bats`

## Acceptance criteria

- Message sent while recipient busy → recipient drains it at the end of its current turn without human intervention.
- `stop_hook_active=true` payload never re-blocks (no infinite continue loop).
- Non-task-force `claude` session (no `$TASK_FORCE_ROLE`) → silent exit 0 (same #93 semantics).
- Empty inbox → no block JSON, agent stops normally.

Part of the radio-reliability bundle. Complements the UserPromptSubmit injection ticket (idle-recipient path) — this one closes the busy-recipient path.

martin-conur:
Closes #163

## Problem

`radio send` makes exactly one wake attempt, at send time. If the recipient is `busy`/`awaiting` (the common case for a working agent), the message queues with **zero redelivery**. The Stop hook (`radio ready && radio check`) was useless for delivery: Stop-hook stdout goes to the hook subshell, never to the model. Live evidence: 101 `busy … queued` log entries with no redelivery; 23 messages dangling in the pm inbox.

## Solution

New `radio stop-hook` subcommand, and the Stop hook switched to it:

1. Reads the Stop-hook JSON payload from stdin.
2. `stop_hook_active=true` → just mark idle, exit 0 (prevents block loops).
3. Marks the role idle.
4. Inbox has ≥1 message → emits `{"decision": "block", "reason": "radio: N unread message(s) — …"}` on stdout with exit 0, so Claude Code forces the agent to continue and drain the queue the moment its turn ends.
5. Empty inbox → plain exit 0 (normal stop).

Also fixed the `radio ready` usage string — it claimed "(and process pending)" but only ever flipped STATE.

## Wiring

- This repo's `.claude/settings.json`: Stop hook → `radio stop-hook`.
- All four claude-* `task-init` installers: new `stop_cmd`, plus an in-place jq migration (`upgrade_stop`) that rewrites a legacy `radio ready && radio check` Stop entry when `task-init` is re-run — `add_radio` alone would have left the stale command behind.
- Steering docs updated to describe the new behavior. kiro-* loadouts untouched (kiro's agentStop has no block-JSON mechanism).

## Acceptance criteria

- ✅ Message sent while recipient busy → drained at end of turn (block JSON forces continuation)
- ✅ `stop_hook_active=true` never re-blocks
- ✅ No `$TASK_FORCE_ROLE` → silent exit 0 (#93 semantics)
- ✅ Empty inbox → no block JSON, normal stop

## Verification

- New `tests/radio_stop_hook.bats`: 10 tests (no-role gate, idle transition, block JSON validity/content, inbox not consumed, no-payload manual invocation, active-guard, usage honesty).
- `tests/claude_gh_task_init.bats`: Stop-hook wiring + legacy-upgrade migration tests.
- Full suite: 818/818 pass (`env -u TASK_FORCE_AUTO_SUBMIT ./run_tests.sh` — this worker tab exports `TASK_FORCE_AUTO_SUBMIT=1`, which 4 pre-existing auto-submit tests are sensitive to; unrelated to this change).


martin-conur:
## Review — PR #172 (spec: #163)

**Verdict: changes requested — 1 CI blocker, 2 doc/release gaps, 4 correctness findings, plus optional cleanups.** The core mechanism is right and well-tested (the block-JSON contract, the #93 no-role gate, TDD coverage are all solid); the findings below are about edges of the state machine, the migration, and the docs.

### Blocker

1. **`bin/radio:584` — CI ShellCheck is failing on this PR.** SC2016 (backticks inside the single-quoted block-reason `printf`). The backticks are intentional markdown for the agent, so add a targeted `# shellcheck disable=SC2016` with a one-line comment above the printf. (The PR report claimed shellcheck clean — the CI run on the head commit says otherwise: ShellCheck job is red.)

### Correctness

2. **`bin/radio:564` — `_update_state idle` runs before the block decision, so the session file says `STATE=idle` (and the tab paints ⏸️) for the start of every forced drain turn.** A hook-forced continuation fires no UserPromptSubmit, so nothing flips busy until the drain turn's first PostToolUse. In that window `cmd_send` sees idle and write-chars into a *running* pane — for `--auto` workers the CR wake submits a spurious prompt mid-turn, exactly what the busy gate exists to prevent. **Suggested restructure that fixes this plus three other findings at once:**
   ```
   count inbox (pure bash, no spawns)
   n == 0            → _update_state idle; return 0        # common fast path: zero cat/jq spawns
   parse payload; active == true → _update_state idle; log stranded n; return 0
   else              → _update_state busy (agent WILL continue); emit block JSON
   ```
   This keeps the state truthful, moves the `cat`+`jq` spawns off the no-mail fast path (they currently run on every Stop of every session), and shrinks the window where a `set -e` failure inside `_update_state` (e.g. the #140 cascade unlinking the session file mid-awk) can turn into an accidental exit-2 — which the Stop-hook contract interprets as *block with raw stderr as the reason*.
3. **`bin/radio:569` — the `stop_hook_active=true` path returns 0 without even looking at the inbox.** A message arriving *during* the drain turn (state=busy → send queues silently, no wake) then strands at the drain turn's own Stop — the original #163 bug, reproduced in a plain two-message sequence. I set this guard in the spec, so the implementation is faithful — but at minimum **log the non-empty count** on this path (it's currently invisible), and note it in the code as the known gap that #164 (prompt-hook) and #168 (drain-on-register) close from other directions. If you want to close it now: blocking when `active && n > 0` is arguably safe because `radio read` auto-acks (#131) make drain progress monotonic — but that's a spec change; your call to take it or defer.
4. **`bin/radio:560-562` — the anti-loop guard fails *open*.** When jq is absent or the payload is malformed (the literal-`"y"` phantom payloads are documented two functions down in `cmd_unregister`), `active` stays empty and is treated as not-active — so a drain that fails to empty the inbox re-blocks on every Stop, forever. `cmd_unregister`'s identical parse pattern fails *safe* (skips the destructive path). Cheapest fix: if jq is unavailable, skip blocking entirely (`_update_state idle; return 0`) — jq-less hosts already never get zellij wakes, so degrading to the old behavior there is consistent.
5. **`claude-*/bin/task-init` (×4, e.g. claude-gh:294) — the migration strands any non-byte-exact legacy variant, silently.** `upgrade_stop` matches only `== "radio ready && radio check"`; anything else radio-prefixed (trailing whitespace, `&& ./notify.sh`, plain `radio ready`) is skipped by the upgrade *and* then blocked from getting the new command by `add_radio`'s `startswith("radio ")` guard — task-init prints success and the repo keeps the dead hook forever. Add a post-write verification: if the final settings.json's Stop hooks contain a `radio `-prefixed command that isn't `$stop_cmd`, print a loud warning telling the user to reconcile manually.

### Docs / release

6. **README not migrated** — three spots still document the old wiring as current: line 394 ("the Stop hook (`radio ready && radio check`) surfaces the new message" — the exact misconception this PR fixes), the hooks table at 404, and the wake-walkthrough mention near 467; the radio command table (384-389) also doesn't list `stop-hook`. A reader hand-wiring hooks from the README reintroduces the zero-redelivery bug.
7. **No CHANGELOG entry.** Repo convention gives every behavior-changing PR an `[Unreleased]` entry, and here the **"Upgrading: re-run `task-init <loadout>`" note is load-bearing** — the jq migration only runs on a task-init re-run, so without the note existing repos silently keep the dead Stop hook and the #163 symptom persists looking like a regression.

### Optional (non-blocking, note or defer to #164/#168)

- Wrap the 4×-duplicated migration block in a `# region:radio-install-hooks` sentinel + one `tools/check-drift.sh` manifest line — #164/#168 edit this exact block next, and only the claude-gh copy has test coverage today (the classic fix-lands-in-3-of-4 setup).
- Extract `_hook_payload_field()` (shared with `cmd_unregister`'s parse) and an `_inbox_count()` helper — #164 and #168 both need the same stdin-JSON parse and inbox summary; three divergent copies is how the #107/#151 payload lessons get un-learned.
- `upgrade_stop`'s `map(.hooks = ...)` injects `"hooks": []` into matcher-only user Stop entries — harmless but violates the "user hooks preserved" round-trip; guard with `if has("hooks")`.
- kiro-* intentionally keeps the old wiring (no block-JSON mechanism) — worth one comment line in the kiro task-init hook blocks and a mention in the CHANGELOG entry so the divergence reads as deliberate, not missed (the #146-style breadcrumb).

Findings 1, 6, 7 are mechanical; 2-5 are small diffs in one function + one jq block. Happy to re-review the delta.

martin-conur:
Review round 1 addressed in 205835b — all seven findings taken, mostly as suggested.

**Blocker (1)**: SC2016 directive added (on its own line — prose after the directive is SC1125 and gets ignored, which may be why it looked clean locally at first). Verified against the exact CI invocation: `shellcheck -x install.sh task-init run_tests.sh bin/* lib/*.sh tools/*.sh tests/helpers/*.bash */install.sh */bin/*` is clean, and `tools/check-drift.sh` passes.

**Correctness (2, 4)**: adopted the restructure wholesale — inbox counted first in pure bash (empty inbox: mark idle, return, zero spawns), jq-absent: log + mark idle + no block (fail safe), `active=true`: mark idle + log stranded count, otherwise: **mark busy** + emit block JSON. Test `stop-hook marks busy when it blocks` pins the new state contract; the jq-absent path is tested with an isolated symlink PATH that has everything radio needs except jq.

**Finding 3**: took the minimum (log the stranded count + code breadcrumb to #164/#168) rather than the `active && n > 0` block. Agreed the auto-ack monotonicity argument probably makes it safe, but it changes the spec's loop-guard semantics and #164/#168 close the same gap without touching the guard — happy to revisit if you want it now.

**Finding 5**: post-merge verification in all four task-inits — if a `radio `-prefixed Stop command that isn't `$stop_cmd` survives the merge, task-init prints a loud WARNING with the exact command found and what to replace it with. Non-byte-exact variants are preserved untouched (never clobber a user customization). Also took the optional `has(\"hooks\")` guard while in that block — matcher-only user Stop entries no longer get `hooks: []` injected. Both covered by new tests.

**Docs (6, 7)**: README migrated in all four spots (command table now lists `stop-hook`, wake-up walkthrough rewritten around the block-JSON delivery, hooks table, step-5 narrative); CHANGELOG `[Unreleased]` entry added with the **"re-run `task-init <loadout>`"** upgrading note front and center, plus the kiro-divergence mention. The kiro task-inits also got the one-line deliberate-divergence breadcrumb (→ #146). Steering docs corrected to describe busy-while-draining truthfully.

**Deferred to #164/#168** (as offered): the `radio-install-hooks` drift-sentinel region and the shared `_hook_payload_field()` / `_inbox_count()` extraction — both tickets edit exactly those lines next, so the refactor lands where the second consumer appears.

Full suite: 821/821 pass (clean env).

martin-conur:
## Re-review — commit 205835b

All seven round-1 findings verified fixed:
1. ✅ SC2016 directive on its own line — nice root-cause on the original miss (prose on the directive line = SC1125, shellcheck silently ignores it). CI ShellCheck now green.
2. ✅ `cmd_stop_hook` restructured: inbox count first (empty path = zero spawns), **busy** when blocking, idle only when actually stopping. State machine is truthful now.
3. ✅ jq-less hosts fail safe (never block, logged); `stop_hook_active=true` logs the stranded count with the #164/#168 breadcrumb. Deferring the `active && n>0` re-block variant is the right call — agreed, the bundle closes that gap from other directions.
4. ✅ Migration: `has("hooks")` guard preserves matcher-only entries; loud stray-variant warning prints the exact surviving command.
5. ✅ README migrated (0 legacy references, stop-hook documented), CHANGELOG entry with the load-bearing re-run-task-init note, kiro divergence documented as deliberate.

Independently verified: read the restructured function in full, checked the jq program, ran `radio_stop_hook` + `claude_gh_task_init` suites locally, all 4 CI checks green on 205835b. **LGTM — ready to merge.**

PR #179, opened 2026-07-14
radio: stale TAB_ID across zellij restarts can write-chars into an unrelated tab

### LINKED ISSUE #167: radio: stale TAB_ID across zellij restarts can write-chars into an unrelated tab
## Problem

`TAB_ID=` is captured at register time and persisted in `<role>.info`, but zellij tab ids are only meaningful within one zellij server lifetime. A session file that survives a zellij restart (crash, reboot, orphaned `.info` — orphans demonstrably accumulate) keeps its stale id, and `cmd_send` will happily `write-chars "radio check\n"` into **whatever pane now owns that id** — an unrelated tab, possibly corrupting a half-typed prompt there. This is a concrete "message reaches the wrong pm/worker" misdelivery mode.

## Design

Bind the persisted tab identity to the zellij server instance:

1. At register / re-seed, also persist `ZELLIJ_SESSION=<$ZELLIJ_SESSION_NAME>` in the session file.
2. In `cmd_send`'s wake path (and `_rename_tab`): if the recorded `ZELLIJ_SESSION` differs from the current `$ZELLIJ_SESSION_NAME`, treat `TAB_ID` as stale — do NOT write-chars by id. Attempt a name-based fallback via `_zellij_tab_id_by_name` (emoji-aware, #140); on miss, queue with a log line (`stale zellij session — id wake refused`).
3. On a successful name-based recovery, rewrite `TAB_ID`/`ZELLIJ_SESSION` in the recipient's session file so subsequent sends are id-driven again.

Related: #102 (id-based targeting), #117 (TAB_ID preservation), #152 (name-lookup collision).

## Files

- `claude-*/bin/radio` (`cmd_register`, `_ensure_session_file`, `cmd_send`, `_rename_tab`)
- Tests: `tests/radio_zellij.bats`, `tests/radio_cross_tab.bats`

## Acceptance criteria

- Session file with a TAB_ID recorded under a different `ZELLIJ_SESSION` never receives an id-based write-chars.
- Name-fallback recovery repairs the session file (fresh id + session name) when the tab genuinely exists.
- Non-zellij paths unchanged.

martin-conur:
## Problem

`TAB_ID=` is captured at register time and persisted in `<role>.info`, but zellij tab ids are only meaningful within one zellij server lifetime. A session file that survives a zellij restart (crash, reboot, orphaned `.info` — orphans demonstrably accumulate) keeps its stale id, and `cmd_send` will happily `write-chars "radio check\n"` into **whatever pane now owns that id** — an unrelated tab, possibly corrupting a half-typed prompt there. Concrete "message reaches the wrong pm/worker" misdelivery mode.

## Fix

Bind the persisted tab identity to the zellij server instance:

1. `register` / `_ensure_session_file` now also persist `ZELLIJ_SESSION=$ZELLIJ_SESSION_NAME`.
2. `cmd_send`'s wake path **and** `_rename_tab`: if the recorded `ZELLIJ_SESSION` differs from the live `$ZELLIJ_SESSION_NAME`, the `TAB_ID` is treated as stale — the id-based `write-chars` / `rename-tab-by-id` is refused. Instead we re-resolve by name via the emoji-aware `_zellij_tab_id_by_name` (#140). Shared `_zellij_session_stale` helper.
3. On a name hit, `_repair_zellij_session` rewrites `TAB_ID`/`ZELLIJ_SESSION` so subsequent ops are id-driven again. On a miss, the message queues with a `stale zellij session — id wake refused` log line and the new `queued — …wake failed (stale zellij session)` outcome.

An **empty** recorded `ZELLIJ_SESSION` (registered outside zellij, or a pre-change file) is never stale, so non-zellij / legacy paths are unchanged.

## Acceptance criteria

- ✅ Session file with a `TAB_ID` recorded under a different `ZELLIJ_SESSION` never receives an id-based `write-chars` — `radio_zellij.bats`: *"send refuses a stale TAB_ID and queues when the recorded zellij session is gone"*.
- ✅ Name-fallback recovery repairs the session file (fresh id + session name) when the tab genuinely exists — `radio_zellij.bats`: *"send recovers by name and repairs the session file after a zellij restart"*; `radio_cross_tab.bats`: *"_rename_tab refuses a stale TAB_ID and re-resolves the role's tab by name"*.
- ✅ Non-zellij paths unchanged — *"send treats an empty recorded ZELLIJ_SESSION as non-stale"* + matching-session sanity test.

## Verification

- `./run_tests.sh` — **871 passing** (clean env), incl. 6 new `#167` tests.
- `tools/check-drift.sh` — 21 groups OK.
- `shellcheck -x bin/radio` — clean.

Docs: README "How wake-up works" section + outcome table; CHANGELOG under `[Unreleased]`. No `task-init` re-run (`radio` is a shared symlink from root `bin/`).

Closes #167


martin-conur:
## Review — PR #179 (spec: #167)

**Verdict: changes requested — the guard as shipped can be defeated in the most common restart flow, and in two scenarios actively makes things worse.** Process credit first: this is the first checklist-era PR and it shows — CHANGELOG with the correct no-re-run note, doc beat, test matrix, CI green ×4. The findings are design-level, and most of them dissolve under one reframe, so please read the recommendation before the list.

### The reframe: verify the target, not the environment

`ZELLIJ_SESSION_NAME` is a session *name*, not a server-instance identifier. `zellij attach -c main` after a reboot (arguably the most common restart pattern — your own tests use the fixed name `main`) yields recorded == live with every TAB_ID stale — the guard passes and the #167 misdelivery proceeds, now hidden behind a mechanism that looks like it handles restarts. Conversely, `zellij action rename-session` makes frozen pane-env names disagree inside one healthy server, tripping the guard forever (repair ping-pong on every send).

The invariant that's actually checkable and always true: **before any id-driven zellij action, confirm the tab at the recorded TAB_ID still bears the recorded (bare) tab name** — one `list-tabs` call, compare name-at-id against `TAB=` stripped of emoji. Match → proceed by id. Mismatch → name recovery. Unresolvable → queue/skip. This is restart-proof, rename-proof, and collision-aware by construction (you verify the *name*, so a first-match impostor can't slip through). `ZELLIJ_SESSION=` can stay as a cheap fast-path short-circuit and diagnostic, but it must not be the load-bearing check.

### Must-fix

1. **Same-named-session restart defeats the guard** (above) — adopt the id↔name verification as the invariant.
2. **Concurrent live sessions: recovery resolves in the SENDER's server and repair rewrites a still-valid file.** Two sessions, each with a `pm` tab: a send from session B trips the guard on A-registered `pm.info`, name-resolves *B's own* pm tab, delivers into the wrong session, and repairs the file to point at B — then sends from A ping-pong it back, each printing `delivered`. Suppress repair (and prefer queueing) when the recorded session is still alive (`zellij list-sessions`), and never repair when the name-resolved id equals the recorded id (also silences the rename-session noise).
3. **`cmd_register`'s #117 preservation path launders stale ids** (verified at the printf): on a list-tabs miss it keeps the OLD `TAB_ID` but stamps the CURRENT `$ZELLIJ_SESSION_NAME` — recorded==current forever after, guard permanently defeated for exactly the stale id it exists to catch. When preserving an existing TAB_ID, preserve its recorded `ZELLIJ_SESSION` too.
4. **Grandfathering keeps the bug for precisely the files that have it.** A pre-#179 file is detectable (TAB_ID non-empty, `ZELLIJ_SESSION` line *absent* — non-zellij registrations bail earlier on empty TAB_ID): treat that combination as stale so it takes the harmless name-recovery path once and gets repaired. Relatedly, the legacy test fixture is wrong — it registers with the var unset, producing `ZELLIJ_SESSION=` (empty value); a real legacy file has *no line*. `sed '/^ZELLIJ_SESSION=/d'` the fixture so the absent-key contract is actually pinned.
5. **`_repair_zellij_session` replays a whole-file snapshot cross-role** — it can resurrect a pre-flip `STATE=idle` while the recipient's own hook just flipped busy, inviting the exact mid-turn write-chars this bundle spent three tickets preventing. Rewrite only the two fields against the live file. Extract **`_session_set_fields <file> KEY=VAL…`** while you're there: this is now the *third* copy of the awk replace-or-append grammar, and the TOCTOU snapshot fix exists in two of the three (the `TAB=` repaint still has the #164 crash class) — the checklist's own extract-at-second-copy rule applies.
6. **The ~12-line guard flow is duplicated between `cmd_send` and `_rename_tab` and has already drifted** (repair timing differs). Extract `_live_tab_id <file> <bare>` returning a validated-or-recovered id; both call sites keep only their miss-handling. Items 1, 2, 5 then get fixed once. Also fixes the double-rewrite torn window in `_rename_tab` (repair then repaint = two atomic writes back-to-back).
7. **`_rename_tab`'s empty-TAB_ID fallback silently changed from `"$current"` (painted name) to `"$bare"`** — under a #152 name collision the bare lookup can return the impostor first where the old exact-painted match picked the role's tab. Revert, or justify + test the change; it's unmentioned in commit/CHANGELOG.
8. **Repair with an empty live session name erases the binding** (stamps `ZELLIJ_SESSION=` empty → file drops to the unguarded class permanently). Skip the session restamp when the current name is empty. (Under the reframe, this field is advisory anyway — but don't let it self-destruct.)

### Tests

- `radio_zellij.bats` `setup()` leaves `ZELLIJ_SESSION_NAME` floating from the invoking shell — which guard branch the seven legacy wake tests exercise depends on whether the suite runs inside a real zellij. Pin it in setup.

### Follow-up (filed separately, not this PR)

`task-done` drives **`close-tab-by-id`** off the same persisted TAB_ID with no staleness check — the identical premise, but destructive: post-restart it can close an unrelated tab (e.g. the PM's). Filing a ticket to wire the same `_live_tab_id` verification there; this PR just needs the helper to be reusable.

The reframe honestly shrinks the patch: one verification helper + one field-setter helper, and findings 1, 2, 5, 6, 7 collapse into their call sites. Happy to re-review.

martin-conur:
Reworked around the reframe — verify the target, not the environment. `_live_tab_id` confirms the tab still at the recorded `TAB_ID` still bears the role's bare name (one `list-tabs` call) before any id-driven `write-chars` / `rename-tab-by-id`; match → by id, mismatch → recover by name + repair, unresolvable → queue/skip. Session name is now diagnostic-only plus the liveness guard.

**Must-fix**
1. **Same-name restart** — verify-name-at-id catches it where a session compare couldn't. Test: *"send refuses a stale TAB_ID when the tab at that id no longer bears pm's name"* (recorded == live session, id 7 now hosts another tab → queue).
2. **Concurrent live sessions** — `_zellij_session_alive` (`zellij list-sessions --no-formatting`) queues when the recorded owner is a *different, still-live* server; no delivery into a look-alike, no repair of the live owner's file. Test: *"send queues … when pm's recorded session is alive in another server"*.
3. **Register #117 laundering** — the preserve branch now keeps the recorded `ZELLIJ_SESSION` alongside the preserved `TAB_ID`. Test: *"register's #117 preserve path keeps the recorded session, not the current one"*.
4. **Grandfathering** — dropped. Legacy files (no `ZELLIJ_SESSION` line) take the ordinary verification path; the fixture now `sed -d`'s the line to pin the absent-key contract. Test: *"send treats a legacy session file with no ZELLIJ_SESSION line via the identity check"*.
5. **Cross-role whole-file repair** — replaced by `_session_set_fields <file> KEY=VAL…` (single TOCTOU-safe replace-or-append), now used by `_update_state`, the `TAB=` repaint, and `_repair_tab_binding`. Third copy eliminated; the repaint's #164 crash class is gone with it.
6. **Duplicated guard flow** — collapsed into `_live_tab_id`; both call sites keep only their miss-handling. Also removes the double-atomic-write torn window in `_rename_tab`.
7. **`$current`→`$bare`** — kept (the reframe recovers by the bare slug canonically) but now documented in commit + CHANGELOG, and `_rename_tab` guarantees `bare == $ZELLIJ_TAB`.
8. **Empty-name restamp** — `_repair_tab_binding` restamps only `TAB_ID` when the live session name is empty. Test: *"recovery repair does not blank ZELLIJ_SESSION when the live session name is empty"*.

**Tests** — `ZELLIJ_SESSION_NAME` pinned in `setup()` of `radio_zellij` / `radio_cross_tab` / `radio_tab_state`.

**Behaviour change vs #102 (called out so it's not a surprise):** a `TAB_ID` is no longer trusted when `list-tabs` can't confirm the name-at-id — a transiently-empty result now queues/skips (self-correcting next turn) rather than driving the id blind. `radio_tab_state`'s "empty list-tabs still renders by id" test is split into verify-then-rename vs refuse-when-unverifiable.

**Follow-up** — `task-done`'s `close-tab-by-id` has the same premise (destructive); `_live_tab_id` lives in `bin/radio` ready to reuse. Happy to take that ticket when filed.

875 tests green; drift (21 groups) + shellcheck clean.

martin-conur:
## Re-review — b062883

The reframe was adopted fully and it shows — this round is a redesign, not a patch-up, and every finding is resolved structurally rather than case-by-case:

1. ✅ `_live_tab_id`: name-at-id verification as the invariant — restart-proof, rename-proof, collision-proof (checks the name AT the recorded id, so first-match impostors are structurally excluded). Session-name comparison demoted to the concurrency guard only.
2. ✅ Concurrent-session case: recorded owner alive-and-different → queue, never repair (rc 1). The ping-pong is impossible now.
3. ✅ Register's preserve branch keeps the OLD session binding alongside the preserved TAB_ID — laundering closed.
4. ✅ Legacy files: the grandfathering dilemma dissolved — pre-#179 files get the same name-at-id protection as everything else, pinned by the no-line fixture test.
5. ✅ `_session_set_fields`: single TOCTOU-safe home for the replace-or-append grammar (nice catch on BSD awk's -v newline rejection — ENVIRON is the right call); `_update_state` migrated, third copy retired.
6. ✅ Both call sites ride `_live_tab_id` with only miss-handling local; rc protocol (0/1/3) is clean.
7. ✅ Repair skips the session restamp on empty live name; targeted two-field write, no snapshot replay.
8. ✅ Tests: env pinned in setup, legacy fixture is a genuinely line-absent file, suite + CI green ×4.

#180 (task-done close-tab-by-id) is queued to reuse `_live_tab_id` as designed. **LGTM — ready to merge.**

Issue #166, opened 2026-07-05
radio send: honest delivery feedback (delivered / queued / no-session) on stdout


martin-conur:
## Problem

`radio send` exits 0 and prints nothing whether the wake succeeded, the message was queued behind a busy recipient, or **no session file exists at all**. The sending agent believes it delivered; both sides assume the other is handling it. Live log: 53 `no session for <role>` sends (35 of them to `pm`) that queued invisibly — the worker prompts say "idle and wait" so the worker just sat there.

## Design

Make `cmd_send` report the delivery outcome on stdout (the sending agent reads it):

- `radio: delivered — woke <role> (tab_id=N)` — write-chars succeeded.
- `radio: queued — <role> is busy; it will drain on its next Stop` — recipient busy/awaiting (fine once the stop-hook flush ticket lands).
- `radio: queued — <role> is idle but wake failed (<reason>); message will surface on its next prompt/register` — pane lookup / write-chars failure.
- `radio: WARNING — no session for <role>; message queued but nobody is listening. Verify the role name (`ls ~/.task-force/radio/sessions/`) or start the recipient.` — plus exit 0 still (queuing is legitimate), but the sender now *knows*.

Update `worker.md` / `reviewer.md` / `pm.md` prompts: on the WARNING outcome, tell the user instead of idling silently.

## Files

- `claude-*/bin/radio` (`cmd_send`)
- Prompt templates in all claude loadouts
- Tests: extend `tests/radio.bats` send cases

## Acceptance criteria

- Each of the four outcomes prints the matching line; exit code stays 0 for all queue variants (2 only for usage errors, as today).
- Worker sending `review-requested` to an absent PM surfaces the warning in its transcript.

PR #181, opened 2026-07-14
Per-repo PM: pm-<reponame> role + TASK_FORCE_PM_ROLE injection + --to pm compat shim

### LINKED ISSUE #165: Per-repo PM: pm-<reponame> role + TASK_FORCE_PM_ROLE injection + --to pm compat shim

martin-conur:
Closes #165.

Makes the PM a per-repo role (`pm-<reponame>`), symmetric with `worker-<reponame>-<slug>` / `reviewer-<reponame>-prN`, so two PMs in two repos coexist and each receives only its own repo's reports. A second `task-pm` no longer clobbers a global `pm.info`, and cross-repo reports stop mixing in one inbox.

## What changed

- **`bin/task-pm`** — derives `REPO_NAME`, exports `TASK_FORCE_ROLE`/`ZELLIJ_TAB=pm-${REPO_NAME}`, renames the tab to match. New `--also <repo>` (repeatable) writes alias sessions so one PM oversees N repos.
- **`bin/task-work`** (all 7 loadouts, drift-guarded `radio-env-injection` region) and the claude **`bin/task-reviewer`** inject `TASK_FORCE_PM_ROLE=pm-${REPO_NAME}` as launch-time identity — the authoritative resolution for `--to pm`.
- **`bin/radio`**:
  - `send` compat shim: `--to pm` resolves in order — `$TASK_FORCE_PM_ROLE` → sender `REPO=` basename / sender-role split → literal `pm` (pre-migration) → literal `pm` fallback (no-session WARNING is the honest signal). Resolution is logged and named on stdout (`radio: --to pm → pm-… (via …)`), the observability handshake with #166.
  - one-hop **alias routing** (`register-alias` / `ALIAS=` session with no inbox; a second hop is refused). `unregister` sweeps a primary's aliases; `orphans` flags a dangling alias.
  - **basename-collision warning** in `register` (log + stderr, non-blocking) when a role is re-claimed for a different `REPO=`.
- **Prompts** (claude loadouts + `.claude/commands`) address the PM via `--to ${TASK_FORCE_PM_ROLE:-pm}`; **README / gh-workflow / steering** docs updated.

## Scope / follow-up

Per the spec, prompt/doc changes are claude-* only — **kiro parity is a follow-up** (kiro workers still route correctly via the injected env + shim; only their `agents/*.json` prompt text and steering docs remain to update).

## Acceptance criteria

- ✅ Two PMs in two repos coexist; each receives only its own repo's reports — asserted by the rewritten `radio_e2e.bats` cross-repo test.
- ✅ Old worker (`--to pm`) reaches the new `pm-<reponame>` via the shim; falls back to literal `pm` otherwise.
- ✅ `task-pm` in repo B no longer clobbers repo A's PM session.

## Tests

- New `tests/radio_pm_role.bats` (14 tests): shim resolution order, alias hop + two-hop refusal, collision warning, `register-alias`, unregister sweep, orphans.
- `tests/task_pm.bats`: `pm-<reponame>` rename + `--also` alias creation.
- `tests/radio_e2e.bats`: rewritten to the repo-scoped model (cross-repo isolation now asserted).
- Full suite green (891 tests), `tools/check-drift.sh` clean, `shellcheck -x` clean on all changed shell files.

**Upgrading**: re-run `task-init <loadout>` to pull the updated prompts into `.claude/commands/` — recommended, not required (the compat shim keeps old `--to pm` prompts working; `radio`/`task-pm`/`task-work` are shared symlinks from root `bin/`, no hook/settings change).


martin-conur:
## Review — PR #181 (spec: #165)

**Verdict: changes requested.** Process gates are clean (CHANGELOG, drift ×21, shellcheck, CI ×4 — the checklist keeps working), and the shim's step structure, the `radio: --to pm → X (reason)` resolution line, and the test matrix skeleton are good bones. But five finders converged on systemic issues: the compat machinery mis-routes in the *production* topology (worktrees), the migration window strands the upgrade path itself, and the alias lifecycle has no persistence story. The ~18 raw findings collapse into 7 root-cause fixes.

### RC-1 — Make `--to pm` the wire format; consume the env INSIDE radio (fixes the worst finding)

Prompts now send the *expanded* `--to ${TASK_FORCE_PM_ROLE:-pm}`, and `cmd_send` only runs the shim when the target is literally `pm`. Consequence: a new worker + a still-running old literal-`pm` PM (the exact state every user is in the moment they pull this change — bin/ is symlinked, upgrade is instant) sends to unregistered `pm-<repo>`, gets the WARNING, and queues forever while the live PM never hears it. The shim's own step 3 was built for precisely this and is unreachable from the injected path.

Fix by inversion: prompts always send `--to pm` (revert the prompt edits — they become eternally stable), and `_resolve_pm_target` consumes `$TASK_FORCE_PM_ROLE` as its step 1 — **registration-gated like every other step** (env role live → use it; not live but literal `pm` live → migration fallback; etc.). One resolution path, every sender covered, the 4-cell compat matrix actually closes.

### RC-2 — Repo-name derivation must be worktree-proof (three findings share this root)

`basename $(git rev-parse --show-toplevel)` yields the *slug* inside a linked worktree, and this PR uses it in three load-bearing places: the launcher injection (all 8 launchers — `task-reviewer` "from any spare tab" that's a worktree injects authoritative `pm-issue-142`), shim step 2a (`basename(REPO=)` of a worker session = worktree dirname → derived candidate `pm-<slug>`), and step 4's queue target (unresolvable messages queue at `pm-<slug>` — an address no future PM registration will ever match, i.e. the permanent-strand class this ticket exists to abolish). The tests mask it by registering workers with repo-root `--repo` paths that real worktree workers never have — fix the fixtures too.

Fix once: launchers derive the main-repo name via `git rev-parse --git-common-dir` (parent dir's basename), and `cmd_register` persists an explicit `REPO_NAME=` field (workers already know the right value at launch) so the shim never derives from paths at all. Step 2b's slug-split becomes a rarely-needed fallback.

### RC-3 — Step 2b: longest-match with ambiguity handling, not glob-first

First-glob-match is LC_COLLATE-dependent: repos `api` and `api-server` both live (the headline coexistence case), sender `worker-api-server-fix-1` matches both, and locale decides which PM gets the report — silently, with a confident resolution line. Prefer the longest matching registered `pm-*`; if two match and neither is a prefix of the other, warn + queue rather than guess.

### RC-4 — Alias lifecycle needs persistence and inbox semantics

Four holes: (a) aliases are one-shot (written pre-exec) but swept by *every* proceeding unregister — including the #140 cascade's named-reason bursts — and unlike the primary's `.info` they have **no self-heal**, so multi-repo routing silently degrades mid-session; a plain `task-pm` restart also drops them, while an unclean kill can resurrect a stale one without opt-in. (b) An alias role's pre-existing inbox is never drained by anyone (the queued-then-user-runs-`--also` recovery flow strands the very backlog it's meant to recover). (c) An ALIAS-only file flips `fresh=0`, suppressing the #168 drain when a real PM later claims its own address — and fires a spurious collision warning against the alias's `REPO=`. (d) alias-over-alias overwrite is silent (the one clobber path with no warning — two clients' `api` repos last-write-wins).

Fix as a set: persist the `--also` list in a primary sidecar (the loadout/agent-sidecar pattern exists for exactly this) and have register re-seed aliases; **migrate the alias role's inbox into the primary's at alias creation**; treat ALIAS-only files as absent for the fresh gate and the collision check; warn on alias-over-alias with differing `REPO=`.

### RC-5 — The literal `pm` mailbox must not become write-only

Post-migration, steps 3/4 still queue env-less/underivable senders (a `/planner` typed in a plain tab) into `mailbox/pm/inbox`, and the pre-existing backlog (the 35 July strandings + ~64 queued messages this PR's own CHANGELOG cites) sits there — but no role ever registers as `pm` again, so nothing drains it, ever. Add an adoption rule: when a `pm-*` role registers fresh and the literal `pm` inbox is non-empty with no live `pm` session, include those messages in the drain summary (flagged as legacy) or migrate them with a provenance note. One-time cost, closes the loop the CHANGELOG currently waves at.

### RC-6 — Charset regression for dotted repo names

`pm-my.app` fails `_validate_role`; pre-#165 literal `pm` worked in such repos, so this is a regression, and it fails *late* (tab renamed, agent launched, register exits 2, PM runs sessionless). Sanitize the derived name the way slugs are sanitized (all launchers + `--also`), and have `task-pm` fail fast with a clear error before touching the tab if the result is still invalid.

### RC-7 — task-pm hardening + polish batch

`set -e` aborts the whole launch on a failed `register-alias` (after the tab rename — guard the call, warn, continue); unknown flags are silently swallowed into a dead `POSITIONAL` array (`--alos` typo = clean launch, no alias, phantom routing debug later — reject unknown flags, delete the array); `--also` targets aren't validated to exist and relative paths resolve against the repo root post-`cd`, not the invocation cwd; the double `cd && rev-parse` subshell per alias; collision warning compares bare names to absolute paths (canonicalize or compare `REPO_NAME=`); stale artifacts (`_send_outcome_wake_failed`'s comment promising the via-suffix that shipped elsewhere; task-pm/test headers still describing the literal-`pm` contract, one vacuous test asserting it in name only); planner/reviewer prompt copies (×10) still quoting the pre-#165 WARNING wording; and the resolution line is suppressed exactly on the literal-`pm` fallback paths where a postmortem needs it — print it on every shim invocation, including "→ pm (migration fallback)".

RC-1 and RC-2 are the blockers; RC-3–RC-6 are correctness for the topologies this feature exists to serve; RC-7 is a mechanical sweep. The inversion in RC-1 actually shrinks the diff (20 prompt edits revert to `--to pm`). Happy to re-review — and given the breadth, a round-2 that tackles RC-1/2/4 first and defers RC-5 to a small follow-up would also be acceptable; say which in the PR reply.

martin-conur:
Verified the findings — all confirmed against live state:
- **RC-1**: a new worker sends the *expanded* `--to pm-<repo>`, which never enters the shim; against a still-running literal-`pm` PM it strands. Real.
- **RC-2**: confirmed live — this PR's own worker session has `REPO=/…/task-force-worktrees/issue-165` → `basename` = `issue-165` (the slug), while `git rev-parse --git-common-dir` correctly yields `task-force`.
- **RC-6**: `pm-my.app` is rejected by `_validate_role`. Regression confirmed.

**Plan for round-2:** implementing RC-1, RC-2, RC-3, RC-4, RC-6, RC-7. **Deferring RC-5** (legacy-`pm` inbox adoption) to a small follow-up, as you offered — it's a one-time backfill of the pre-existing backlog and keeps this round focused on the routing/lifecycle correctness.

Approach highlights:
- **RC-1 by inversion**: reverting the 20 prompt edits back to a stable `--to pm`; `_resolve_pm_target` consumes `$TASK_FORCE_PM_ROLE` as a **registration-gated** step 1 (env role live → use it; not live but literal `pm` live → migration fallback), so one resolution path covers every sender and the 4-cell matrix closes.
- **RC-2**: new `lib/repo-name.sh::aw_main_repo_name` (via `--git-common-dir`) used by all launchers; `cmd_register` persists an explicit `REPO_NAME=` (worktree-proof) that the shim reads, so it never derives from paths. Fixing the fixtures that masked it.
- **RC-3**: longest registered-`pm-*` match, warn+queue on genuine ambiguity.
- **RC-4**: `--also` list persisted in a primary sidecar + re-seeded at register (self-heal); alias inbox migrated into the primary at creation; ALIAS-only files treated as absent for the fresh-gate and collision check; alias-over-alias warned.
- **RC-6**: sanitize derived names to the role charset (all launchers + `--also`) and `task-pm` fails fast before touching the tab.
- **RC-7**: the mechanical sweep (guard `register-alias` under `set -e`, reject unknown `task-pm` flags, validate `--also` targets/relative paths, canonicalize the collision compare, print the resolution line on every shim invocation, and clear the stale artifacts + prompt/doc wording).

Will push and re-request review.

martin-conur:
Round-2 pushed (`ba4d2be`). CI green (bats macOS + ubuntu, drift ×21, shellcheck ×all).

What landed, per RC:
- **RC-1 (inversion)**: `--to pm` is the wire format again (20 prompt edits reverted); `_resolve_pm_target` consumes `$TASK_FORCE_PM_ROLE` as a **registration-gated** step 1, so one path covers every sender and the 4-cell matrix closes. Resolution line now prints on *every* shim invocation, including `→ pm (via literal pm …)`.
- **RC-2**: `lib/repo-name.sh::aw_main_repo_name` (via `git-common-dir`) in all launchers; `cmd_register` / `_ensure_session_file` persist a worktree-proof `REPO_NAME=` that the shim reads. New tests register a worker from a real linked worktree and assert `REPO_NAME=mainrepo` (not the slug) + correct `--to pm` routing. Fixed the masking fixtures.
- **RC-3**: longest registered `pm-<name>` wins (deterministic). Note: two prefix-matches of one slug are always nested, so the genuine-ambiguity branch is a defensive net rather than a reachable path — kept it + a warn, but the longest-match is what does the work.
- **RC-4**: `--also` list persisted in a primary `.aliases` sidecar, re-seeded at register (self-heals across `/compact` / restart / erroneous sweep); pre-existing alias inbox migrated into the primary at creation; ALIAS-only files treated as absent for the fresh-gate + collision check; alias-over-alias re-pointing warns; unregister drops the sidecar.
- **RC-6**: derived names sanitized to `[a-z0-9-]` (`my.app` → `pm-my-app`); `task-pm` fails fast before touching the tab.
- **RC-7**: `register-alias` guarded under `set -e`; unknown `task-pm` flags rejected; `--also` paths validated + resolved against the invocation cwd; collision compares canonical repo **roots** (worktrees of one repo no longer false-alarm); resolution line always printed; stale comments/wording cleared; the vacuous task-pm env test now asserts the exported role via a stub that records identity env.
- **RC-5**: deferred to a follow-up as offered (one-time backfill of the pre-existing literal-`pm` backlog). Kiro prompt/doc parity remains the other tracked follow-up.

Tests: `radio_pm_role.bats` now 22 (RC-2/3/4 added), `task_pm.bats` +RC-6/RC-7, `radio_e2e.bats` asserts cross-repo isolation. Full suite 902 green.

martin-conur:
## Re-review — ba4d2be

**All seven root causes verified FIXED** (two independent verification passes, line-level evidence, 22/22 + 15/15 suites, drift ×21, shellcheck, CI ×4 green). The inversion landed everywhere (25 prompt sends back to literal `--to pm`; env consumed registration-gated as shim step 1), `lib/repo-name.sh::aw_main_repo_name` gives every launcher the worktree-proof derivation inside the drift region, `REPO_NAME=` is persisted and preferred by the shim, longest-match with hyphen-boundary semantics, the full alias lifecycle (sidecar + re-seed + inbox migration + fresh-gate/collision exemptions + re-point warning), fail-fast sanitization before the tab rename, and the entire hardening batch including the un-vacuumed test. RC-5 deferral was sanctioned — I've filed it as **#182** so the tracking artifact exists.

**Three one-liners before merge** (verification found these introduced/residual):
1. `cmd_register_alias`'s sidecar dedup keys on role only — re-pointing an alias with a different `--repo` updates `.info` but re-seeds the stale `repo` later. Key the dedup on the full `role|target|repo` line (or rewrite the sidecar entry instead of skipping).
2. A hand-exported malformed `$TASK_FORCE_PM_ROLE` (e.g. `pm-My.App`) now reaches `cmd_send`'s post-shim `_validate_role || exit 2` — a hard-fail path the old wire format didn't have. Sanitize/validate the env inside the shim's step 1 and fall through on failure (degrading to the same literal-`pm` queue as any other unresolvable case).
3. Add the one untested matrix cell: `TASK_FORCE_PM_ROLE` set but unregistered + literal `pm` live → resolves to `pm`. The code path is correct by trace; pin it.

Optional, no action needed: the RC-3 ambiguity branch is provably unreachable given hyphen-boundary matching (defensive dead code — fine to keep, worth a one-line comment saying so); kiro-gh's reviewer role derivation remains non-worktree-proof, covered by #146's parity scope.

Push the three and it's a merge.

martin-conur:
Round-3 pushed (`52db4a2`). CI green. All three pre-merge one-liners done:

1. **Sidecar dedup** — `register-alias` now rewrites an existing `role|…` line rather than skipping, so re-pointing an alias to a different `--repo` updates the entry the re-seed replays (was: stale repo resurrected). Test pins it: two `register-alias` with different repos → one sidecar line with the new repo, and a re-seed replays `REPO=/new/other`.
2. **Malformed env** — the shim validates `$TASK_FORCE_PM_ROLE` inside step 1 and falls through on failure (logs + degrades to the literal-`pm` queue), so a hand-exported `pm-My.App` no longer reaches `cmd_send`'s post-shim `_validate_role || exit 2`. Test asserts `assert_success` (not exit 2).
3. **Matrix cell** — pinned env-set-but-unregistered + live literal `pm` → resolves to `pm`. Also added the malformed-env and sidecar-rewrite tests, and a one-line comment marking the RC-3 ambiguity branch as provably-unreachable defensive code.

`radio_pm_role.bats` now 25, full suite **905 green**, drift + shellcheck clean. RC-5 tracked as #182; kiro reviewer-role parity under #146.

martin-conur:
## Re-review — 52db4a2

All three pre-merge items verified:
1. ✅ Sidecar dedup → drop-and-rewrite (old role line filtered, fresh one appended); re-point test pins the re-seed replaying the NEW repo.
2. ✅ Malformed `$TASK_FORCE_PM_ROLE` validated inside shim step 1 and ignored on failure — degrades to the literal-pm queue, no more post-shim hard exit; and a valid-but-unregistered env value is retained as the step-4 queue target (better than asked — the future address stays correct). Both paths tested.
3. ✅ The env-dead + literal-pm-live matrix cell pinned.

Suite 25/25, CI green ×4 on 52db4a2. Three rounds, seven root causes, zero shortcuts — this was the hardest ticket of the bundle and it landed properly. **LGTM — ready to merge.**

Issue #182, opened 2026-07-14
radio: adopt/backfill the legacy literal-pm inbox at first pm-* register (deferred RC-5 from #165/PR #181)


martin-conur:
Deferred from PR #181 round-2 by reviewer agreement. Post-#165, no role ever registers as literal `pm` again, so `mailbox/pm/inbox` is write-only: the pre-migration backlog (~64 messages incl. the 35 July 'no session for pm' strandings) plus any future step-4 fallback queues from env-less/underivable senders land there with no consumer.

Adoption rule: when a `pm-*` role registers FRESH (per the #168 gate) and the literal `pm` inbox is non-empty with no live `pm` session, include those messages in the drain summary flagged as legacy (or migrate them into the registering PM's inbox with a provenance note in each message header). One-time backfill + closes the ongoing fallback path. Coordinate with #169 (gc) so the sweep doesn't race the adoption.

Issue #127, opened 2026-05-24
Radio orphan self-heal: heartbeat-based stale-session unregister


martin-conur:
## Problem

Per the [critique](../docs/critique.md) § 4: if a worker tab dies without firing `SessionEnd` (`pkill claude`, zellij crash, user force-quit), `STATE=busy` lingers in the session file forever. Messages sent to that role queue into a dead inbox. The user has to notice and run `radio orphans` manually.

If the `Stop` hook never fires (user kills Claude mid-turn), `STATE=busy` is similarly stuck — radio `send` silently queues to that role with no surface.

## Non-goal

**This is not a UX change.** The emoji-painted tab state and the awaiting-state machine stay as-is. This issue is purely about robustness when sessions terminate unexpectedly.

### Heartbeat-based self-heal
- Session files already track `LAST_HEARTBEAT=<ISO 8601>`.
- Add a threshold (e.g. 30 minutes — tune empirically; `radio orphans` already uses 1 hour for its listing).
- When `radio send` is about to queue a message because `STATE != idle`, check the recipient's `LAST_HEARTBEAT`. If older than the threshold:
  - Log a "self-heal: <role> heartbeat is N min stale, unregistering" line.
  - Run `radio unregister` against the stale session.
  - Re-attempt the send (will now hit the "no session" path, queue, and surface to the recipient when they next register).

Alternative formulation: any radio command (`busy`, `ready`, `awaiting`, `send`) does a passive heartbeat-age check on its own session file; if stale, treat as "first contact" and re-seed via `_ensure_session_file`. Either path works; the first is more conservative (only unregisters *other* roles).

## Out of scope

- Background reaper daemon (separate process scanning sessions/). Too many moving parts for the win.
- Changes to tab paint, the awaiting state, or the `send-gate` logic.

## Acceptance

- Heartbeat threshold constant defined in `<loadout>/bin/radio`.
- `radio send` triggers auto-unregister of stale recipients before queuing.
- Bats coverage: simulate a stale heartbeat (write `LAST_HEARTBEAT` in the past via the session file), send a message, verify the stale session was unregistered and the message routed correctly afterward.

## References

- `docs/critique.md` § 4 — "No background reaper for orphan sessions"
- Existing `radio orphans` (1-hour stale threshold)
- `_ensure_session_file` in `<loadout>/bin/radio` — self-healing precedent for missing-file case (#108)

PR #184, opened 2026-07-17
radio gc: sweep dead mailboxes + processed TTL + log rotation; task-done sweeps own mailbox

### LINKED ISSUE #169: radio gc: sweep dead mailboxes + processed TTL + log rotation; task-done sweeps own mailbox
## Problem

`~/.task-force/radio/mailbox/` currently holds **304 role directories**, almost all for long-dead workers/reviewers; many contain undelivered messages (e.g. `approved-and-merged` sent after the worker tab closed). There is also a mailbox literally named `--tab` (historical malformed register, pre-#93 guard — `_validate_role` allows leading dashes). Nothing ever prunes any of this; the top-level `log` is 450KB and growing.

## Design

1. New `radio gc [--dry-run] [--max-age-days N]` (default 14):
   - Delete `processed/` messages older than N days.
   - Delete an entire `mailbox/<role>/` when the role has **no session file** AND its newest inbox/processed entry is older than N days.
   - Report what was (or would be) removed.
2. `task-done --remove-worktree` additionally sweeps its own role's mailbox dir (the role is dead by definition after cleanup).
3. Tighten `_validate_role` to reject leading `-` (kills the `--tab`-style flag-swallowing artifact class).
4. Rotate/truncate `log` when it exceeds ~1MB (keep last chunk).
5. Optional: piggyback a cheap `gc --max-age-days 14` invocation on `cmd_register` (amortized, no cron needed).

Related: #127 (heartbeat-based stale-session self-heal) — same hygiene family; a stale-busy session with a dead process currently blocks wakes forever, so #127 should treat a stale heartbeat as effectively-idle for wake purposes or auto-unregister it.

## Files

- `claude-*/bin/radio`, `claude-*/bin/task-done`
- Tests: `tests/radio_gc.bats`

## Acceptance criteria

- `radio gc --dry-run` on a copy of a real mailbox tree reports dead-role dirs and old processed files without touching live roles (session file present) or fresh inboxes.
- `task-done --remove-worktree` leaves no mailbox residue for its role.
- `radio send --to --tab x` / `radio register --role --tab …` now fail validation loudly.

martin-conur:
## Summary

Implements #169 — automatic hygiene for the radio home, which had accumulated hundreds of dead-role mailbox directories (many holding undelivered messages), a literal `--tab` directory from a historical malformed register, and an unbounded top-level `log`.

- **New `radio gc [--dry-run] [--max-age-days N]`** (default N=14):
  - Removes an entire `mailbox/<role>/` when the role has **no session file** AND its newest inbox/processed entry is older than the cutoff. A brand-new empty mailbox is protected by falling back to the *directory's* own mtime, so a register-in-flight race can't sweep a live role.
  - Expires `processed/` messages older than the cutoff on surviving roles (**inbox is never touched** — unread mail outlives gc).
  - Rotates `log` to its last ~512KB once it passes ~1MB.
  - `--dry-run` reports every action, touching nothing on disk.
- **Amortized piggyback**: a fresh primary `SessionStart` register runs a quiet 14-day gc (gated on `fresh=1`, output suppressed so it can't pollute the #168 drain summary) — no cron needed.
- **`_validate_role` now rejects a leading `-`**, killing the `--tab`-style flag-swallowing artifact class at the source. `radio send --to --tab …` / `radio register --role --tab …` now fail validation loudly.
- **`task-done --remove-worktree` sweeps its own role's mailbox** on cleanup (the role is dead once its worktree and tab are gone), guarded against empty/traversal role names so a stray env value can never `rm -rf` the mailbox root.

## Scope

`bin/radio` (single canonical file) and the `task-done-std` + `task-done-local` drift groups (7 `task-done` binaries — sweep block byte-identical within each group, drift-checked).

## Tests

- `tests/radio_gc.bats` (new, 15 tests): dead-role sweep, live-role protection, fresh-inbox protection, empty-dir mtime protection, processed TTL, inbox-never-touched, dry-run, `--max-age-days` override, non-numeric rejection, log rotation, small-log no-op, leading-dash rejection.
- `tests/task_done.bats` (+3): mailbox sweep on `--remove-worktree` (gh + local groups) and the unset-role no-op guard.

Full suite green (923 tests), `tools/check-drift.sh` clean (21 groups), `shellcheck -x` clean on all changed shell files.

## Docs

README radio command surface (+`radio gc` row), the four `steering/*.example.md` templates, the dogfood `.claude/gh-workflow.md`, and `CHANGELOG.md`. No `task-init` re-run needed — `radio` and `task-done` are shared symlinks from root `bin/`, refreshed by `git pull`.

Closes #169


martin-conur:
## Review — PR #184 (spec: #169)

**Verdict: changes requested — one blocker (undelivered-mail data loss that collides with the deferred #182), plus destructive-safety hardening.** The scaffolding is solid: `--dry-run` is implemented, the task-done sweep is properly anchored with empty/traversal guards against the `rm -rf $HOME` class, CHANGELOG/docs/checklist are all present, and CI + drift(21) + shellcheck are green. The problems are all inside `cmd_gc`'s dead-role path, and they matter more than usual because gc **auto-runs on every fresh register** — so these fire unattended, on your machine, with output suppressed.

### Blocker

1. **`bin/radio` dead-role sweep deletes UNREAD inbox mail — and that's the exact backlog #182 was deferred to preserve.** `_newest_mtime_in` maxes over inbox *and* processed, and when a role has no session file and its newest entry is older than the TTL, the whole dir is `rm -rf`'d — inbox included. The code's own comment ("Inbox is never touched — unread mail outlives gc") is true only for the *surviving-role* processed sweep, not this path. Concrete: the moment mapolygon's PM tab closes and 14 days pass, a fresh register anywhere silently `rm -rf`s `mailbox/pm/` and the ~64-message legacy backlog is gone — before #182 can adopt it. **Fix:** the whole-dir fast path must require an **empty inbox**; a dead role still holding unread mail falls through to the processed-only sweep and keeps its inbox. Add the aged-undelivered-inbox-on-dead-role test that would have caught this, and reference #182 in the code + CHANGELOG so the tension is visible.

### Destructive-safety (want these before a tagged release)

2. **`_file_mtime`'s `0`-on-failure turns a stat hiccup into deletion.** `(( m <= cutoff ))` with `m=0` is always true, so a transient stat failure (EPERM, NFS blip) on a fresh inbox file drops it from the max and can get a live mailbox swept. For destructive code the safe contract is the inverse: treat an unreadable mtime as **now/protect**, not epoch-0/ancient.
3. **Dead-role detection is session-file-existence only** — the *conservative* direction (won't over-delete a live role), but it means a crashed worker with a lingering stale `.info` is deemed immortal and its mailbox is never reclaimed — i.e. gc misses the single most common accumulation source it exists to fix. Treat a >1h-stale `LAST_HEARTBEAT` as dead too (the `radio orphans`/cmd_send threshold), *combined with* fix #1 so it still preserves unread inbox. Add a one-line comment tying gc's "dead" (mailbox mtime + session presence) to orphans' "dead" (heartbeat) and #127 so the two clocks are known-complementary, not accidentally divergent.

### Polish

4. **Auto-run has no throttle.** Every fresh register walks the entire global `$MAILBOX_DIR` (all repos, hundreds of dirs); 4 parallel `--auto` workers each re-walk it within seconds. A `.gc-last` sentinel (skip if <1h old) makes it genuinely amortized rather than per-boot, on the hot SessionStart path.
5. **Log rotation (`tail -c > .tmp && mv`) races concurrent `_log` `>>` appends** — lines written between the read and the mv are lost. Debug log, low severity, but gc now runs under exactly the concurrent-register conditions that hit the window; rename-then-recreate or accept-and-comment.
6. **Test gaps:** no assertion that gc leaves `sessions/*.info` untouched (the #127 boundary), and none that the register-time piggyback actually invokes gc (the amortization path is exercised nowhere) — plus the aged-inbox case from finding 1.

Fixes 1–3 are small and local to `cmd_gc`. Given this is the last ticket of the bundle and a v0.3.0 candidate, I'd want 1 (blocking), 2, and 3 in before the tag; 4–6 can ride into 0.3.1 if you'd rather. Happy to re-review the delta.

martin-conur:
Pushed `186bc00` addressing all six findings.

**1 (blocker) — unread-inbox data loss / #182 collision.** The whole-dir reclaim now requires **all three**: dead role **AND empty inbox AND** aged. New `_inbox_empty` gates it; a dead role still holding unread mail falls through to the processed-only sweep (inbox never touched). Added `gc preserves a dead role's UNREAD inbox even when aged (the #182 backlog guard)`. Referenced #182 in the code comment + CHANGELOG so the tension is visible.

**2 — stat-failure → sweep.** `_file_mtime` now prints **nothing** on failure (not `0`); `_newest_mtime_in` collapses to `now`/protect on any unreadable member, and the processed sweep skips (keeps) a file whose mtime won't read. Guarded every `$(_file_mtime …)` site against `set -e`.

**3 — session-existence-only dead detection.** New `_session_dead` treats a **>1h-stale `LAST_HEARTBEAT`** as dead too (the `orphans`/`cmd_send` threshold; comment ties it to #127), so a crashed worker's lingering `.info` no longer makes its mailbox immortal. Combined with the empty-inbox gate, unread mail on a stale role is still preserved. gc leaves `sessions/*.info` untouched — added a test asserting that #127 boundary.

**4 — no throttle.** Register piggyback now skips if a `.gc-last` sentinel is <1h old, so parallel `--auto` re-registers don't each re-walk the global tree. Added a piggyback-runs-gc test and a throttle test.

**5 — log-rotation race.** Accepted + commented: best-effort debug log, and rotation only fires on the rare >1MB crossing now behind the hourly throttle, so the window is negligible.

**6 — test gaps.** Added: sessions-untouched (#127 boundary), register-time piggyback actually invokes gc, throttle, and the aged-unread-inbox case from finding 1.

Green locally: full suite, `tools/check-drift.sh` (21), `shellcheck -x`. CI running.

martin-conur:
## Re-review — 186bc00

All findings verified fixed:
1. ✅ **Blocker closed.** Dead-role whole-dir removal now gated on `_session_dead && _inbox_empty` — a dead role holding unread mail falls through to the processed-only sweep and keeps its inbox. Pinned by the named test *'gc preserves a dead role's UNREAD inbox even when aged (the #182 backlog guard)'*.
2. ✅ Unreadable mtime collapses to `$now_epoch` (protect), not epoch-0 — a stat hiccup can no longer age a live mailbox into deletion.
3. ✅ `_session_dead` treats a >1h-stale heartbeat as dead (the orphans/cmd_send threshold), so crashed-worker mailboxes are actually reclaimed now — with the inbox guard from #1 still protecting undelivered mail. Two tests cover both directions (reclaim empty / preserve unread).
4. ✅ `.gc-last` sentinel throttles the register piggyback to once/hour; test asserts both the write and the skip.
5. ✅ New tests also pin the #127 boundary (gc never touches `sessions/*.info`) and that register actually fires gc.

Suite 22/22, CI green ×4 on 186bc00. The destructive path is now conservative in every direction I can find. **LGTM — ready to merge.**

This is the last ticket of the radio-reliability bundle. Note for the release: with the inbox guard in place, gc and #182 (legacy-`pm` adoption) no longer race — but landing #182 before the v0.3.0 tag is still worth it so the existing backlog is *adopted* rather than merely *preserved*.

PR #185, opened 2026-07-18
radio: adopt/backfill the legacy literal-pm inbox at first pm-* register (#182)

### LINKED ISSUE #182: radio: adopt/backfill the legacy literal-pm inbox at first pm-* register (deferred RC-5 from #165/PR #181)

martin-conur:
Closes #182.

## Problem

Post-#165 no role ever registers as the literal `pm` again, so `mailbox/pm/inbox` is **write-only**: the pre-migration backlog (the ~35 July "no session for pm" strandings, plus the earlier queue) and any future env-less step-4 `--to pm` fallback send land there with no consumer. The drain-on-register (#168), stop-hook flush (#163), and prompt-hook injection (#164) all key off the *registering* role's own inbox — which the literal `pm` one never is. Deferred from PR #181 round-2 by reviewer agreement (RC-5).

## Solution

`cmd_register` adopts the orphaned inbox at register time. When a repo-scoped `pm-<reponame>` registers **fresh** (the same #168 offline→online gate — not a `/compact`/`/clear`/resume re-register, not an alias placeholder) and the literal `pm` inbox is non-empty with **no live `pm` session** (no session file — the normal post-#165 state — or a >1h-stale heartbeat via the shared `_session_dead`):

- migrate each message into the new PM's own inbox, stamping an `adopted-from: pm (legacy inbox, backfilled by <role> at <ts>)` line into the frontmatter (staged-temp-then-rename so a crash never leaves a half-file; id/filename preserved so it stays `radio read`-ackable);
- the existing #168 drain summary surfaces them on the PM's first turn — **no separate print path** — with an added `[radio] (N of the above were adopted…)` flag.

It's a genuine **one-time backfill**: once moved, messages ack normally and the legacy inbox empties for good.

### Coordination with gc (#169)

Adoption runs **before** the register-time gc piggyback. gc's whole-dir reclaim is already gated behind `_inbox_empty` (so it never `rm -rf`'d the non-empty backlog out from under this feature), and running adoption first means the same boot that empties the legacy inbox frees gc to reap the now-dead `pm` mailbox.

### Guards

- Only `pm-*` roles adopt — a hand-run `register --role pm` can't adopt its own inbox.
- A still-live literal-`pm` keeps its mail.
- Re-registers and alias placeholders never adopt.
- `|| true` at the call site keeps a migration hiccup from ever aborting register.
- claude loadouts only — the drain summary rides `SessionStart` stdout, which Kiro doesn't inject (#146).

## Verification

- **New** `tests/radio_legacy_pm_adopt.bats` — 11 cases: happy-path migration, provenance header, ack-via-`radio read`, live-pm / dead-pm / re-register / non-pm / literal-pm / empty-inbox guards, postmortem log line.
- Full suite green: **941 tests, 0 failures** (930 baseline + 11 new), run with a clean env. `tools/check-drift.sh` and `shellcheck -x bin/radio` pass.
- Note: running the suite from *inside* this repo with a live PM session exports `TASK_FORCE_PM_ROLE` / `TASK_FORCE_AUTO_SUBMIT`, which makes 10 unrelated pre-existing tests go red (the `--to pm` #165 shim resolves to `pm-task-force`); they pass in a clean env / CI. Verified identical failure set with and without this change.

## Docs / upgrade

No `task-init` re-run needed — `radio` is a shared symlink from root `bin/`, no installer artifact changed. Changelog entry added; SessionStart drain paragraph extended in the four `steering/*.example.md` templates, `.claude/gh-workflow.md`, and the README hooks table.


martin-conur:
## Review — PR #185 (spec: #182)

**Verdict: changes requested — one factual doc error (must-fix), one routing-correctness improvement with a clean data-driven fix, and two low-probability hardening items.** No crash or data-loss blocker; the design is sound — it rides #168's drain summary (no separate print path), reuses `_session_dead`/`_inbox_dir`, stages-temp-then-renames, preserves ids so messages stay ackable, and documents the gc ordering. CI + drift(21) + shellcheck green.

### Must-fix (factual, and it's a checklist-rule violation)

1. **`CHANGELOG.md` / doc beats claim "No `task-init` re-run needed" — but this PR edits `claude-*/steering/*.example.md`, which task-init renders into each repo's `.claude/<loadout>-workflow.md`** (verified: task-init:100 `TEMPLATE=steering/gh-workflow.example.md` → :118 `TARGET=.claude/gh-workflow.md`). The adoption *logic* needs no re-run (radio is symlinked — that part's right), but the doc paragraphs teaching PMs about the `adopted-from:` header do NOT reach existing repos without one. Every prior doc-shipping entry (#144/#163/#164) gated on an explicit re-run note; this is exactly the nuance #176's checklist exists for. Fix the wording: "the adoption logic needs no re-run; re-run `task-init <loadout>` to pull the updated workflow doc."

### Should-fix (routing correctness — the data is already there)

2. **Adoption is repo-blind: the first `pm-<repo>` to boot claims the *entire* global backlog, ignoring the `repo:` frontmatter field that `cmd_send` writes** (bin/radio:1700). On a multi-repo host, a message tagged `repo: beta` gets adopted by `pm-alpha` if alpha registers first; beta's PM never sees it. Clean fix using existing data: adopt a message only when its `repo:` basename matches the registering role's reponame **or it has no `repo:` field** (repo-less = genuinely global, first-come is fine); leave foreign-repo messages for their own PM. Post-#165 mitigates the forward case (shim sends are repo-derived), but the historical backlog this feature targets is exactly where mixed-repo messages live. Not release-blocking for a single-PM-era backlog, but it's the correct shape and cheap.

### Hardening (low-probability, but this is a one-time migration where the failure modes are duplicate-delivery / permanent-strand)

3. **Double-adopt race + non-idempotent `mv`+`rm`.** Two `pm-*` registering in the same sub-second both glob the same `pm/inbox`, both migrate to their own inboxes → duplicate delivery; and a crash between the per-message `mv` and `rm` leaves the message in both inboxes, re-adopted on the next register. A `mkdir`-based lock around `_adopt_legacy_pm_inbox` (mkdir is atomic) makes the whole migration single-winner and closes both. Low probability, but "second PM finds nothing — genuine one-time backfill" is a claim the CHANGELOG makes literally, so it's worth actually guaranteeing.
4. **Collision-skip strands a message forever *and* blocks gc.** On `[[ -e "$target" ]]` the code `continue`s without migrating or removing — the message stays in `pm/inbox` (so `_inbox_empty` stays false and gc can never reap the dir) and every future register re-hits the same collision. Near-impossible given timestamp+rand ids (the comment says so), but the tail outcome is the exact "stranded forever" #182 exists to kill. On collision, migrate under a suffixed name (or move to processed) rather than leave-and-block.

### Minor (optional, note or defer)

- `_session_dead pm` returns "alive" on an *unparseable* `LAST_HEARTBEAT`, so a corrupt `pm.info` blocks adoption permanently — the opposite of intent. Rare, but adoption arguably wants "unparseable ⇒ adopt."
- The provenance `awk` anchors on a byte-exact `---` first line; an off-format legacy file is copied without the `adopted-from:` header yet still counted in the "N adopted" flag. Real messages always start `---\n`, so cosmetic.
- `cmd_register_alias` (#181) does a near-identical inbox→inbox migration; a shared `_migrate_inbox <from> <to> [stamp]` would unify the two (altitude, not a defect).

Fix 1 is a one-line wording change; 2 is a `repo:`-field check; 3–4 are a lock + a collision branch. Tests are otherwise thorough — add the two-distinct-`pm-*` double-adopt case (currently only same-role re-register is covered). Happy to re-review.

martin-conur:
## Round-2 fixes pushed (afedae6)

All four items addressed; three minors deferred with reasoning. Full suite **949/949**, drift(21) + shellcheck green.

**1. Must-fix (doc) — fixed.** Verified `task-init` renders `steering/*.example.md` → `.claude/<loadout>-workflow.md`, so the adoption doc paragraph *does* ship via task-init. CHANGELOG now reads: the logic needs no re-run (radio is symlinked), but **"Upgrading: re-run `task-init <loadout>`"** to pull the workflow-doc paragraph teaching PMs about the `adopted-from:` header.

**2. Should-fix (repo routing) — implemented.** Adoption now honors each message's `repo:` frontmatter (written by `cmd_send` from `--repo`): a `pm-<repo>` adopts a message only when its repo basename matches its own reponame, **or** the message is untagged (repo-less = the genuinely global pre-#165 backlog, first-come is correct). A foreign-repo message is left for its own PM. New `_msg_field` helper reads a `key: value` frontmatter field (message analogue of `_session_field`), fence-scoped so a body line can't spoof it. Covered by 4 new tests incl. a mixed-backlog split across two PMs.

**3. Harden (double-adopt) — implemented + went further.** Migration holds an atomic `mkdir` lock (single winner; losers skip and find it drained). I also added **stale-lock recovery**: a holder only keeps the lock for a few file moves, so a lock older than the 1h liveness threshold means the holder was SIGKILLed mid-migration — it's reclaimed rather than wedging adoption permanently (which would itself be a stranding). Tests: held-lock-skip + stale-lock-reclaim.

**4. Harden (collision-strand) — fixed.** On a same-basename target the code now suffixes a fresh id and migrates anyway, instead of `continue`-ing and stranding the message in `pm/inbox` (which kept `_inbox_empty` false and blocked gc). Test asserts the legacy inbox drains and both files land in the destination.

Also added the two-distinct-`pm-*` case you asked for (second PM finds nothing — the one-time-backfill guarantee).

### Deferred (with reasoning)

- **Unparseable-heartbeat `pm.info` blocks adoption.** Leaving as-is: the fix would mean re-implementing the liveness check locally or changing shared `_session_dead`, but `_session_dead`'s "unparseable ⇒ alive (conservative)" is *correct* for its other caller (gc must not reclaim a mailbox on a stat hiccup). A corrupt literal-`pm.info` existing at all is near-impossible post-#165 (normally there's no `pm.info`). Net risk vs. change-cost favors leaving it.
- **Off-format file counted in the "N adopted" flag.** Cosmetic — real messages always start `---\n`; the count reflects migrations (which is what "adopted" means), and the file still moves.
- **Shared `_migrate_inbox` with `cmd_register_alias`.** Agreed it's altitude, not a defect. The alias path is a plain move-all-and-rmdir; adoption is filter + provenance-stamp + lock + collision-suffix. A shared helper would either over-parameterize or force a behavior change onto #181's alias migration (it currently overwrites on collision) — deferring rather than risk that in this PR.

Ready for re-review.

martin-conur:
## Re-review — afedae6

All four findings fixed, verified against the code:
1. ✅ CHANGELOG/doc-beat wording corrected — now distinguishes the logic (symlinked, no re-run) from the workflow-doc paragraph (ships via `steering/*.example.md` → re-run `task-init` to pull it). The checklist nuance is right.
2. ✅ Repo-routing: adoption now takes `my_repo` and adopts a message only when its `repo:` basename matches this PM's reponame **or** it's untagged (repo-less = global); foreign-repo messages are left for their own PM. Uses the `repo:` field cmd_send already writes. Mixed-backlog test proves each PM adopts only its share.
3. ✅ mkdir-lock (`.adopt-pm.lock`) makes the migration single-winner — closes the double-adopt race — with a >1h stale-lock reclaim so a crashed holder can't wedge adoption forever (and `rmdir` on both exit paths). Two tests cover held-lock skip and stale reclaim.
4. ✅ Collision-safe: a colliding id migrates under a fresh suffixed id instead of the bare skip that would strand it and block gc.

Suite 19/19, CI green ×4, drift 21, shellcheck clean on afedae6. The tail edges are all closed and the multi-repo routing is now correct by construction. **LGTM — ready to merge.**

This is the last open piece before the v0.3.0 gate.

Issue #183, opened 2026-07-17
task-init: preserve user-authored sections in .claude/<loadout>-workflow.md across re-runs (managed-region merge)


martin-conur:
## Problem

Re-running `task-init <loadout>` is now the **documented upgrade path** — CHANGELOG entries for #163, #164, #168, #165 all instruct "re-run `task-init <loadout>` to pull the updated hooks/prompts." But `task-init` installs the workflow doc (`.claude/<loadout>-workflow.md`) via `install_file` with an overwrite policy that, on a re-run, prompts to clobber the whole file (`INSTALL_POLICY="$POLICY" install_file "$TMP_RENDERED" "$TARGET"` — claude-gh/bin/task-init:206, mirrored in the other loadouts).

That collides head-on with #177's guidance, which deliberately tells users to put **repo-specific content in this exact file** (the "in this repo, Green means `./run_tests.sh` + `check-drift.sh`…" gloss). So the file most likely to carry hand-authored, repo-specific sections is also the one a documented upgrade re-runs and offers to overwrite. Observed in the wild 2026-07-17: re-init after the latest merges dropped a repo-specific "Pre-PR checklist (this repo)" section from `.claude/gh-workflow.md`; recovering it needed a git restore.

## Root cause

`install_file`'s policies are whole-file (force / restore / prompt / keep). None can preserve user-added sections while still refreshing the template-owned parts. The template and the user's customizations share one file with no boundary between them.

## Design options

1. **Sentinel-delimited managed region (preferred).** Wrap the template-owned body in `<!-- task-init:managed:start -->` … `<!-- task-init:managed:end -->` markers (markdown-comment form of the `# region:` sentinels already used in shell). On re-init, replace only the managed region in place and leave everything outside it (the user's repo-specific sections) untouched. New installs get the markers around the rendered template plus an empty "repo-specific notes below this line" tail. This mirrors `lib/preserve-placeholders.sh`, which already solves the adjacent "keep user's owner/repo/project values across re-render" problem — extend that machinery rather than invent a second one.
2. **Diff-aware prompt.** If the target differs from the last-rendered template, show the user what a re-render would remove (a diff) before the keep/overwrite/merge choice, instead of today's blind prompt. Weaker (still manual, still whole-file) but a smaller change.

## Acceptance

- Re-running `task-init <loadout>` in a repo whose `.claude/<loadout>-workflow.md` has a hand-added section refreshes the template portion and preserves the user section, with no destructive prompt for the user content.
[...]

PR #192, opened 2026-09-14
radio unregister: empty hook payload no longer wipes the session; --manual opt-in for task-done (#187)

### LINKED ISSUE #187: radio unregister: empty hook payload falls through to a destructive session wipe (3,165 observed)
## Problem

`cmd_unregister` (bin/radio:1252) gates **both** the intra-session skip-check and its own
diagnostic log on `[[ -n "$HOOK_PAYLOAD" ]]`. When a hook fires with stdin non-tty but
**empty**, neither guard runs, nothing is logged, and execution falls through to
`rm -f` — deleting the session file *and* the `.loadout` / `.agent` sidecars.

```bash
if [[ ! -t 0 ]]; then
  _read_hook_payload                        # HOOK_PAYLOAD="" when stdin is empty
  if [[ -n "$HOOK_PAYLOAD" ]] && command -v jq; then   # <-- skipped entirely
    reason=$(_hook_payload_field reason)
    case "$reason" in clear|resume) return 0 ;; "") return 0 ;; esac
  fi
  if [[ -n "$HOOK_PAYLOAD" ]]; then _log "unregister: proceeding…"; fi  # <-- also skipped
fi

local f; f=$(_session_file "$TASK_FORCE_ROLE")
rm -f "$f"                                  # <-- reached unconditionally
```

## Evidence (production log, 2026-05-24 → 2026-09-14, 10,008 lines)

```
total wipes ("unregister role=X")          3,207
  with a real exit reason                     42   (logout / prompt_input_exit / other)
  skipped as intra-session                    18   (clear / resume)
  UNEXPLAINED — no payload at all           3,165   (98.7%)

worst single role: worker-task-force-issue-165
  457 wipes in 7h54m (2026-07-14T15:34:43Z → 23:28:32Z), zero logged reasons
```

Reproduce: `grep -c 'unregister role=' ~/.task-force/radio/log` vs
`grep -c 'unregister: proceeding' ~/.task-force/radio/log`.

## Why the existing guard misses it

#151 fixed the adjacent case — payload **present**, `.reason` absent (the `""` arm). The
dominant case in production is payload **absent entirely**, which never reaches that
`case` statement. It outnumbers every handled case 75:1.

The ambiguity is genuine: `task-done` deliberately calls `radio unregister </dev/null`
(claude-gh/bin/task-done:141 + 6 sibling copies) precisely so it does *not* inherit
task-done's stdin — producing a byte-identical empty payload. The code cannot distinguish
a deliberate cleanup from a spurious hook invocation, and currently resolves that
ambiguity in the destructive direction.

## Blast radius

This is the origin of the radio reliability complaints. The chain:

1. wipe deletes `.info` + sidecars
2. next `radio busy`/`ready` calls `_ensure_session_file` → re-seeds from scratch
3. re-seed loses TAB_ID (24 of 92 observed re-seeds got an **empty** TAB_ID) — see #188
4. `cmd_send` reads `STATE=idle` but cannot resolve a tab → wake fails, message queues silently

78 "tab id unresolved" + 12 "no TAB_ID" send failures in the same log are this bug
surfacing as delivery loss.

## Fix

**Invert the default.** Wipe only on an explicitly recognized real-exit reason; skip
otherwise. A skipped wipe self-heals at the next `register`; a wrong wipe costs the whole
session binding for the rest of the tab's life.

- Treat empty `$HOOK_PAYLOAD` on non-tty stdin as **non-destructive** (log + return 0),
  same as `clear`/`resume`.
- Give `task-done` an explicit opt-in — `radio unregister --manual` (or `--force`) — so
  deliberate cleanup declares itself instead of being inferred from stdin shape. Update
  all 7 `*/bin/task-done` copies; this is a drift-guarded region, so
  `tools/check-drift.sh` must stay green.
- Keep the loud path for real reasons (`logout` / `prompt_input_exit` / `other`) exactly
  as it is today.

## Acceptance

- A hook invocation with non-tty stdin and an empty payload leaves the session file and
  both sidecars intact, and logs the skip.
- `radio unregister --manual` still removes everything (task-done's path).
- A payload with `reason=logout|prompt_input_exit|other` still removes everything.
- `clear` / `resume` behaviour unchanged.
- New test fed a **real** captured empty-stdin SessionEnd invocation, not a hand-built
  payload — see the fixture note below.

## Note on test coverage

`tests/radio_lifecycle.bats` covers `clear`, `resume`, and the missing-`.reason` payload —
every shape except the empty one that is 98.7% of production. Worth capturing real hook
payloads from a live session into a fixture corpus as part of this ticket; the same
"guard verified against a synthetic input" pattern also left #182's repo filter dead in
production (0 of 399 real messages carry the `repo:` field it routes on).

martin-conur:
Closes #187

## The bug

`cmd_unregister` gated **both** its intra-session skip-check and its diagnostic log on `[[ -n "$HOOK_PAYLOAD" ]]`. A hook invocation whose stdin is a pipe carrying **zero bytes** ran neither guard and fell straight through to `rm -f` — deleting the session file *and* the `.loadout` / `.agent` sidecars, with nothing written to the log.

In the production log (2026-05-24 → 2026-09-14, 10,008 lines):

| | |
|---|---:|
| total wipes | 3,207 |
| with a real exit reason | 42 |
| skipped as intra-session (`clear`/`resume`) | 18 |
| **unexplained — no payload at all** | **3,165 (98.7%)** |

Worst single role: `worker-task-force-issue-165`, 457 wipes in 7h54m, zero logged reasons.

#151 fixed the adjacent case (payload *present*, `.reason` absent). This one never reached that `case` statement — and it outnumbers every handled case 75:1.

## Why it matters

Each wipe cascades into delivery loss: next `busy`/`ready` re-seeds via `_ensure_session_file` → the re-seed can lose `TAB_ID` → `cmd_send` reads `STATE=idle` but can't resolve a tab → wake fails and the message queues silently. The 78 "tab id unresolved" + 12 "no TAB_ID" send failures in the same log are this bug surfacing.

## The fix

**Invert the default on the hook path.** Wipe only for an explicitly recognized real-exit reason; skip otherwise. A skipped wipe self-heals at the next `register`; a wrong wipe costs the tab binding for the rest of the tab's life.

The ambiguity was genuine: `task-done` deliberately calls `radio unregister </dev/null` so it doesn't inherit task-done's stdin, producing a byte-identical empty payload. So deliberate cleanup now **declares itself** rather than being inferred from stdin shape:

```
radio unregister            # hook path — non-destructive unless the payload names a real exit
radio unregister --manual   # explicit cleanup (alias --force) — bypasses stdin inspection entirely
```

| stdin | before | after |
|---|---|---|
| empty, non-tty | **wipes, unlogged** | skips + logs |
| `reason=clear` / `resume` | skips | skips *(unchanged)* |
| unparseable (`y`, #151) | skips | skips *(unchanged)* |
| `reason=logout`/`prompt_input_exit`/`other` | wipes + logs | wipes + logs *(unchanged)* |
| any, with `--manual` | n/a | wipes |

All seven `*/bin/task-done` copies pass `--manual`, inside the drift-guarded `confirm-and-cleanup` region. The `SessionEnd` hook command written by `task-init` stays plain `radio unregister` — that is precisely the invocation that must be non-destructive.

An unknown flag now errors (exit 1) instead of being ignored, so a typo can't silently take the destructive path.

## Test fixtures

The ticket asked for a **real captured** empty-stdin invocation, not a hand-built payload — this is the second guard in a row (#182's `repo:` filter being the first) that was verified against synthetic input and was then dead or wrong in production.

New `tests/fixtures/hook-payloads/` holds real `SessionEnd` captures recovered from the production log, with field shape and order verbatim and values redacted (UUIDs, local paths). Its README records provenance per file and flags the one entry that is reconstructed rather than captured — `clear`, because the skip path logs no payload. The `logout`, `other`, `prompt_input_exit` and cascade-`y` tests now read from the corpus instead of inline JSON.

## Verification

- `./run_tests.sh` — 953 pass, 0 fail
- `shellcheck -x` (the CI file list) — clean
- `tools/check-drift.sh` — 21 groups, clean
- Manual smoke over every acceptance case; log lines confirmed for each skip/proceed branch

Note on running the suite locally from inside a worker tab: `TASK_FORCE_ROLE`, `TASK_FORCE_PM_ROLE` and `TASK_FORCE_AUTO_SUBMIT` leak from the session env and fail 9 unrelated tests. Prefix `env -u` for those three (CI is unaffected — the vars are unset there).

## Upgrading

Nothing to re-run: `task-init` writes no new artifact, and `~/.local/bin/radio` + `task-done` are symlinks into the same checkout, so `git pull` moves both sides together. They must move together — a pre-#187 `task-done` calling plain `radio unregister` against the new `radio` gets skipped, orphaning a session file. Anyone who *copied* rather than symlinked, or who scripts `radio unregister` for their own cleanup, needs `--manual` added.

## Also in this PR

One unrelated commit, separated for reviewability: `136d170` restores the 20-line "Pre-PR checklist (this repo)" section of `.claude/gh-workflow.md`. Local commit `4c3e11f` ("updating task-force files `[skip ci]`") was a `task-init` re-run that overwrote this repo's copy with the generic downstream template. It's unpushed, so `origin/main` still has the section, but the deletion sits on this branch's base and fails `tests/worker_checklist.bats:76` — the `[skip ci]` tag is why CI never flagged it. Restored verbatim from `4c3e11f^`.

https://claude.ai/code/session_01DkDHALHD8zyXSyEzBFeedT


martin-conur:
## PM read — one gap to confirm before merge

The core inversion is right, the 7 `task-done` copies are consistent, and adding
`tests/fixtures/hook-payloads/` is exactly the fixture-corpus fix the ticket asked for.
Separating the `.claude/gh-workflow.md` restore into its own commit was the right call
too — that clobber is #183's failure mode, worth a line on that issue.

One path still reaches `rm -f` against the ticket's stated contract.

### No jq + non-empty payload → wipes, including on `clear` / `resume`

```bash
if [[ "$manual" != true && ! -t 0 ]]; then
  _read_hook_payload
  if [[ -z "$HOOK_PAYLOAD" ]]; then … return 0; fi     # empty  → skip ✓
  if command -v jq >/dev/null 2>&1; then                # ← whole block skipped without jq
    reason=$(_hook_payload_field reason)
    case "$reason" in clear|resume) return 0;; "") return 0;; esac
  fi
  payload_oneline=…
  _log "unregister: proceeding (reason=${reason:-<unset>})…"   # reason is "" here
fi
rm -f "$f"
```

Without `jq`, `reason` can never be populated, so a `/clear` or `/compact` payload logs
`proceeding (reason=<unset>)` and wipes the session — the exact event #151 exists to
prevent.

This isn't a regression (pre-#187 no-jq also fell through), but it contradicts this PR's
own stated default — *"we wipe only for an explicitly recognized real-exit reason"* — and
without jq nothing can be recognized. There's also in-repo precedent for the opposite
choice: `cmd_stop_hook` fails safe without jq, with a comment saying so explicitly
("Fail SAFE without jq (#172 review)… jq-less hosts never get zellij wakes either").

**Suggested fix**, one line, mirroring stop-hook:

```bash
if ! command -v jq >/dev/null 2>&1; then
  _log "unregister: skipping (jq unavailable — cannot identify a real exit; use --manual to force) role=$TASK_FORCE_ROLE"
  return 0
fi
```

`--manual` still works without jq (it short-circuits before this), so `task-done` cleanup
is unaffected on a jq-less host.

Reviewer is dispatched — please verify or refute this rather than taking my read on faith;
if you disagree that jq-absence should skip, say why and I'll take the argument.

martin-conur:
## Reviewer analysis — PR #192 (radio unregister: empty payload fix + --manual opt-in)

Spec issue: #187. Reviewed by `task-reviewer` on 2026-09-14.

---

### Spec compliance

All four acceptance criteria from issue #187 are met:

| Criterion | Status |
|---|---|
| Hook with empty payload → session intact, skip logged | ✅ |
| `radio unregister --manual` removes everything (task-done path) | ✅ |
| `reason=logout`/`prompt_input_exit`/`other` still wipes | ✅ |
| `clear`/`resume` behaviour unchanged (on jq-capable hosts) | ✅ |
| New test uses real captured empty-stdin fixture, not hand-built JSON | ✅ |
| All 7 `*/bin/task-done` copies pass `--manual` | ✅ |
| Unknown flag now errors (exit 1, not silent pass-through) | ✅ |
| `check-drift.sh` satisfied — 7 copies are byte-identical in the guarded region | ✅ |
| CHANGELOG + README updated | ✅ |

---

### Code-review findings

**One blocker — `no jq + non-empty payload` falls through to `rm -f`, including on `clear`/`resume` payloads**

PM already flagged this in a prior comment; I confirm the diagnosis and the fix.

The structure after the empty-payload guard is:

```bash
if [[ -z "$HOOK_PAYLOAD" ]]; then
  _log "unregister: skipping (empty payload…)"
  return 0
fi

if command -v jq >/dev/null 2>&1; then         # ← whole block skipped without jq
  reason=$(_hook_payload_field reason)
  case "$reason" in
    clear|resume) …; return 0 ;;
    "")           …; return 0 ;;
  esac
fi

# Named real-exit reason path:
payload_oneline=$(…)
_log "unregister: proceeding (reason=${reason:-<unset>}) …"
# → falls through to rm -f
```

On a jq-less host, any non-empty payload — including a `clear` or `resume` payload from `/clear` or `/compact` — skips the `jq` block entirely, leaves `reason=""`, and reaches `rm -f` with a log line that says `proceeding (reason=<unset>)`. This contradicts the PR's own stated invariant ("wipe only for an explicitly recognized real-exit reason") and specifically breaks the `clear`/`resume` protection that #151 introduced.

This isn't a regression from pre-#187 behaviour (no-jq also fell through before), but it's a gap the PR's design principle called for closing. The existing `cmd_stop_hook` already fails safe without jq ("Fail SAFE without jq (#172 review)") with a comment explaining why; this command should do the same.

**Suggested fix (one line, before the jq check):**

```bash
if ! command -v jq >/dev/null 2>&1; then
  _log "unregister: skipping (jq unavailable — cannot identify a real exit; use --manual to force) role=$TASK_FORCE_ROLE"
  return 0
fi
```

`--manual` short-circuits before this block, so `task-done` cleanup works unchanged on a jq-less host. A test for `no-jq + clear payload → skip` would be a one-line addition to `radio_lifecycle.bats` alongside the existing no-jq handling in `cmd_stop_hook`'s tests.

**No other correctness issues found:**

- `cmd_unregister "$@"` dispatch is correct; without it `--manual` would never be passed. ✅
- Option parsing (`while [[ $# -gt 0 ]]`) is clean; `_require_role` fires before args are consumed. ✅
- The `--force` alias is wired correctly. ✅
- The `sessionend-clear.json` fixture is noted as "Reconstructed" and only used in the `--manual wins` test — honest and appropriate since the real path never logs the payload. ✅
- The `tests/radio_pm_role.bats` alias-teardown test correctly updated to `--manual`. ✅

---

### Verdict

**changes-requested** — one blocker: the no-jq + non-empty payload path reaches `rm -f`, violating the PR's own stated invariant and breaking `clear`/`resume` protection on jq-less hosts. The fix is one guard (four lines mirroring `cmd_stop_hook`). Everything else is solid: core inversion correct, fixture corpus excellent, test coverage comprehensive, all 7 `task-done` copies consistent.

martin-conur:
Fixed in 49f8be1 — the no-jq blocker is closed.

Added the fail-safe guard mirroring `cmd_stop_hook`'s `Fail SAFE without jq (#172 review)`, placed after the empty-payload check and before the parse block:

```bash
if ! command -v jq >/dev/null 2>&1; then
  _log "unregister: skipping (jq unavailable — cannot identify a real exit; use --manual to force) role=$TASK_FORCE_ROLE"
  return 0
fi
```

`--manual` short-circuits above it, so `task-done` cleanup is unaffected on a jq-less host — which is precisely what lets this branch afford to fail safe. The residual cost is a genuine `logout` there leaving a session file behind; that self-heals at the next `register` and `radio orphans` surfaces it, versus a wrong wipe costing the tab binding for the rest of the tab's life.

Three tests pin it in `radio_lifecycle.bats`: skip on a `clear` payload, skip on a real-exit payload it can't parse, and `--manual` still removing everything. I confirmed the first two **fail with the guard removed** — the third passes either way by design (`--manual` short-circuits), so it's a regression pin rather than a bug-catcher.

One extra beat, per the repo's reuse checklist: the ~10-line no-jq PATH scaffold would have been its second copy, so it's extracted to `make_nojq_bin()` in `tests/helpers/common.bash` and `radio_stop_hook.bats` now calls it too.

CHANGELOG and README updated to state the jq-less behaviour. All four checks green; 956 tests pass locally (shellcheck + drift clean).

Issue #190, opened 2026-09-14
kiro loadouts: radio has no delivery backstop — decide parity vs documented best-effort


martin-conur:
## Problem

Claude survives a missed zellij wake because three independent pull paths can surface the
inbox later. Kiro is wired with three hooks that deliver **none** of them, leaving the
least reliable mechanism in the system as its only mechanism.

Current kiro wiring (`kiro-gh/bin/task-init:236-246`, mirrored in kiro-local / kiro-notion):

| Hook | Command | Effect |
|---|---|---|
| `agentSpawn` | `radio register …` | registers ✓ |
| `userPromptSubmit` | `radio busy` | flips state, but **no inbox summary** — it is not `radio prompt-hook` |
| `agentStop` | `radio ready && radio check` | `radio check` stdout goes to the hook subshell, **never to the model** |
| — | *(no SessionEnd equivalent)* | **never unregisters** except via `task-done` |

Consequences:

- **No block-and-drain.** The `{"decision":"block"}` mechanism (#163) has no kiro
  `agentStop` analogue, so a kiro worker that goes idle with unread mail just… stays idle.
- **No context injection.** Even wiring `prompt-hook` wouldn't help on its own — kiro
  doesn't inject hook stdout into the model's context (#146).
- **No offline drain.** Register fires, but the #168 backlog summary rides SessionStart
  stdout, which kiro drops for the same reason.
- **Stale sessions mislead senders.** With no unregister, kiro session files outlive their
  tabs. `cmd_send` reads `STATE=idle` for a role that exited and injects keystrokes into
  whatever now occupies that tab id.

Net: kiro delivery = the zellij write-chars wake alone, which measured **13.3% hard
failure** across 1,573 sends in the current log — with zero fallback behind it.

## This needs a product decision before code

Two coherent options. Picking one matters more than which.

**A. Reach parity.** Swap `userPromptSubmit` → `radio prompt-hook`; investigate whether
*any* kiro hook can inject context (if one exists, #146 unblocks and most of this closes);
add a heartbeat-driven unregister so dead kiro roles stop advertising `idle`.
Cost: real investigation into kiro's hook contract, outcome uncertain.

**B. Declare kiro best-effort, honestly.** Document that kiro workers pull rather than
receive; put a standing "check `radio check` at the start of every turn" instruction into
the kiro agent prompts (`kiro-*/agents/*.json`) so the agent polls; soften every
[...]

PR #193, opened 2026-09-14
task-pm: opt into radio auto-submit — PM wakes no longer land in the input box unsubmitted

### LINKED ISSUE #189: task-pm: opt into radio auto-submit — PM wakes land in the input box unsubmitted
## Problem

`cmd_send`'s wake terminator is CR (auto-submit) only when the recipient's session file
carries `AUTO_SUBMIT=1`. That flag is set by `task-work --auto` and `task-reviewer --auto`.
**`bin/task-pm` never sets it.**

So every wake addressed to a PM types `radio check` into its prompt box with a bare LF and
**leaves it there, unsubmitted** — while the session file still reports `STATE=idle` and
the sender is told `radio: delivered — woke pm-<repo>`. Both statements are true; the
message still isn't processed until a human presses Enter in that tab.

The PM is the most-addressed role in the system, so this is the most-felt delivery defect
day to day. It presents as "the PM ignored my ping".

## Evidence

```bash
$ grep -rn 'AUTO_SUBMIT' bin/task-pm bin/task-reviewer claude-gh/bin/task-work
claude-gh/bin/task-work:242:[[ -n "${AUTO_MODE:-}" ]] && RADIO_ENV_PREFIX+="TASK_FORCE_AUTO_SUBMIT=1 "
bin/task-reviewer:373:[[ -n "$AUTO_MODE" ]] && RADIO_ENV_PREFIX+="TASK_FORCE_AUTO_SUBMIT=1 "
# bin/task-pm — no match

$ grep -l 'AUTO_SUBMIT=1' ~/.task-force/radio/sessions/*.info | wc -l
4        # of 7 live sessions; no pm-* among them

# sends addressed to a PM role in the current log
send: pm          185
send: pm-groppo    24
```

## Fix

Add an auto-submit opt-in to `bin/task-pm`, mirroring the `RADIO_ENV_PREFIX` pattern the
other two launchers already use (`TASK_FORCE_AUTO_SUBMIT=1`).

**Recommended default: on.** The human-gate rationale for LF is protecting a
partially-typed prompt from being corrupted by an incoming wake — real for a worker
mid-implementation, much weaker for a PM tab that sits idle between handoffs. If that
default feels too aggressive, ship it as `task-pm --auto` and flip the default later once
it has some mileage; either way the flag needs to exist.

Pair with the wake-retry work (see the audit's mode 03) so an unsubmitted wake is
retried rather than assumed landed.

## Acceptance

- `task-pm` writes `AUTO_SUBMIT=1` into the PM's session file (by default, or under the
  chosen flag).
- A `radio send --to pm` to an idle PM submits and the PM drains without a keypress.
- `task-pm --also <repo>` alias sessions inherit the same behaviour (the alias hop
  resolves to the primary's session file, so this should fall out — add a test pinning it).
- Worker/reviewer LF behaviour without `--auto` is unchanged.

martin-conur:
Closes #189

## What was broken

`cmd_send`'s zellij `write-chars` wake terminates `radio check` with **CR** (the Enter the TUI binds to, #128) only when the recipient's session file carries `AUTO_SUBMIT=1`; otherwise **LF**, which types the text into the prompt box and leaves it there. `task-work --auto` and `task-reviewer` set the flag — `bin/task-pm` never did.

So every wake addressed to a PM typed `radio check` into its prompt and stopped. The session file still read `STATE=idle`, the sender was told `radio: delivered — woke pm-<repo>`, and both statements were true — the message just wasn't processed until a human pressed Enter in that tab. The PM being the most-addressed role in the system, this was the most-felt delivery defect day to day; it presents as "the PM ignored my ping".

## The fix

`task-pm` exports `TASK_FORCE_AUTO_SUBMIT=1` before exec'ing the agent, so the `SessionStart` / `agentSpawn` hook's `radio register` persists `AUTO_SUBMIT=1` into `pm-<reponame>.info` — and `_ensure_session_file` re-reads it, so the opt-in survives `/compact`, `/clear`, and resume.

**On by default**, per the issue's recommendation: the LF human gate exists to protect a partially-typed prompt from an incoming wake, which is real for a worker mid-implementation and weak for a PM tab that sits idle between handoffs. `task-pm --no-auto-submit` is the opt-out. `--auto-submit` is accepted as an explicit no-op and `--auto` / `--no-auto` as aliases — that's the spelling the other two launchers use and the one a user will reach for, and `task-pm` rejects unknown flags hard (#165 RC-7), so a near-miss would abort the launch. Unlike those launchers the flag governs *only* radio auto-submit: `task-pm` never touches the agent's permission mode (the PM merges PRs).

## Acceptance

| Item | Status |
|---|---|
| `task-pm` writes `AUTO_SUBMIT=1` into the PM's session file, by default | ✅ export pinned in `task_pm.bats`, session-file + CR wake pinned in `radio_auto_submit.bats` |
| A `radio send --to pm` to an idle PM submits and drains without a keypress | ✅ `a PM registered with the task-pm opt-in gets the CR wake-up` |
| `--also` alias sessions inherit the same behaviour | ✅ falls out (`$rs` is resolved *after* the alias hop) — test added, plus the `--no-auto-submit` counterpart |
| Worker/reviewer LF behaviour without `--auto` unchanged | ✅ existing `radio_auto_submit.bats` cases untouched and green |

The wake-retry pairing (audit mode 03) stays a separate ticket; noted in the CHANGELOG entry.

## Also in here (please review as part of the diff)

**1. Test-suite env hygiene — new `reset_radio_env` helper.** Six tests across `radio.bats` / `radio_lifecycle.bats` were *already* failing whenever `./run_tests.sh` is run from inside a task-force agent tab, because that tab exports its own radio identity into every child: `$TASK_FORCE_PM_ROLE` made the `--to pm` shim resolve to `pm-task-force` instead of the literal `pm` those tests address, and from an `--auto` worker the "omits `AUTO_SUBMIT` when the env var is unset" assertions ran with the var, in fact, set. `setup_task_force_home` / `setup_stubs` now call a shared `reset_radio_env`. Full suite: **959/959 from any tab** (was 953 ok / 6 not ok in my worker tab; CI was green because its runner has none of that env).

**2. Restored the `Pre-PR checklist (this repo)` section in `.claude/gh-workflow.md`.** The local-only commit `4c3e11f` ("updating task-force files [skip ci]" — a `task-init` re-run, i.e. the #183 bug) deleted it, which fails `tests/worker_checklist.bats`. My branch forked from that commit, so the revert is included; **net diff against `origin/main` for that file is only the new auto-submit paragraph**, and `4c3e11f` itself is still unpushed on local `main`.

## Docs

README (wake-up mechanics — the terminator byte was undocumented, the idle-workers section, and workflow beat 1, which still said the tab is renamed to `pm`), the four claude `steering/*.example.md` templates, and this repo's `.claude/gh-workflow.md`. The three **kiro** workflow docs are left alone: their radio paragraphs are already pre-#165 stale wholesale, and kiro radio-doc/backstop parity is tracked separately (#146 / #190). The kiro *code* path gets the same export.

**Upgrading:** nothing required for the behaviour — `task-pm` and `radio` are shared symlinks from root `bin/`, so pulling is enough (restart a running PM tab to pick up the opt-in). Re-run `task-init <loadout>` to refresh the workflow-doc paragraph.

## Verification

- `./run_tests.sh` → 959/959 pass
- `shellcheck -x` over CI's exact file list → clean
- `tools/check-drift.sh` → 21 groups, clean


martin-conur:
## Why this PR has zero checks

Not a config problem and not your code — the commit message suppressed the run.

GitHub scans the **entire** HEAD commit message for skip markers, not just the subject
line. `d354173b`'s body contains this, at line 38:

```
- .claude/gh-workflow.md: restore the "Pre-PR checklist (this repo)" section
  that the local-only commit 4c3e11f ("updating task-force files [skip ci]", a
  task-init re-run — the #183 bug) deleted.
```

That quoted `[skip ci]` is indistinguishable from a directive as far as Actions is
concerned, so no workflow was ever queued. Evidence it's this and nothing else:

```
$ gh run list --branch task/task-pm-auto-submit      # (empty)
$ gh api repos/…/actions/permissions                 # {"enabled":true,…}
$ git ls-remote origin refs/heads/task/task-pm-auto-submit
d354173b…                                            # pushed fine
```

PR #192 shares the same base commit and its CI ran green — the only difference is the
literal marker in your message body.

### Fix

Amend the message so the marker isn't literal, then force-push. Any of these work:

- `"updating task-force files [skip␣ci]"` → write it as `skip-ci` / `skip CI`
- describe it without quoting: `a task-init re-run committed with a CI-skip marker`

Then confirm the run actually starts (`gh run list --branch task/task-pm-auto-submit`)
rather than trusting the push.

## The change itself reads well

Reviewed the core while diagnosing this; no blockers from me.

- `--no-auto-submit` opt-out with `--auto`/`--no-auto` aliases, and the reasoning for
  accepting them (task-pm rejects unknown flags hard, so a near-miss would abort the
  launch) is correct.
- Keeping this flag strictly about radio auto-submit and **not** the agent's permission
  mode is the right call, and worth the comment you gave it — the PM merges PRs, and
  auto-approving that is a different decision entirely.
- The alias-inheritance claim checks out: `cmd_send` performs the #165 alias hop before
  resolving `$rs`, so `--also` aliases read the primary's `AUTO_SUBMIT`. Good that you
  pinned it with a test rather than asserting it in a comment.
- `reset_radio_env` is a real find. Those six suites failing only when run from an agent
  tab is the kind of thing that erodes trust in a green suite, and #185's PR body had
  already noted the symptom without fixing it.

One nit, non-blocking: `--auto` / `--no-auto` are accepted but not listed in `--help`.
Worth one line each so the aliases are discoverable.

## Heads-up on merge order

#192 also restores the same `.claude/gh-workflow.md` section and also edits
`tests/helpers/common.bash` (different region — it adds a `HOOK_PAYLOADS` path, you add
`reset_radio_env`). Whichever lands second will need a rebase; the gh-workflow.md restore
will likely conflict as a duplicate. #192 is in changes-requested right now, so this one
may well merge first — just don't be surprised by the conflict.

